PR #51, opened 2018-11-01
227 jenkins config

julienmachon:


julienmachon:
That's the Jenkins file nicely setup. Need to be merged to `nexus-web` branch to see if the multi-stage Dockerfile is going to work with our Openshift/Jenkins/DnD setup :man_shrugging:

PR #53, opened 2018-11-02
create header component

julienmachon:
fixes BlueBrain/nexus#241

screenshot from 2018-11-02 16-42-22


src/server/html.ts
```diff
@@ -14,7 +15,9 @@ const html = ({ body, helmet }: { body: string, helmet: HelmetData }): string =>
       ${helmet.title.toString()}
       ${helmet.meta.toString()}
       ${helmet.link.toString()}
+      <link rel="shortcut icon" type="image/png" href="${icon}"/>
       <link rel="stylesheet" href="${base}public/bundle.css" />
+      <meta name="viewport" content="width=device-width, initial-scale=1">
```
mdupont: Nitpicking, but grouping <link> and <meta> tags together makes it better organized (with the exception of the charset <meta> tag which should come before the <title> if we need to have it.

So `${helmet.meta.toString()}` could be moved below the latest <link> and above the viewport <meta>

src/shared/components/Header/Header.less
```diff
@@ -1,7 +1,54 @@
 .Header {
-  background-color: #ff6666;
+  display: flex;
+  justify-content: space-between;
+  height: 60px;
+  background-color: #f0f0f0;
```
mdupont: variables should be preferred to plain values

julienmachon:
Agreed! I just need to figure out how to make them work with my webpack setup 😬

src/shared/components/Header/Header.less
julienmachon: Agreed! I just need to figure out how to make them work with my webpack setup 😬

src/server/html.ts
julienmachon: Fair enough, I like things to be tidy as well

PR #54, opened 2018-11-06
Cookie jwt

julienmachon:
## What is this?
This is a POC to show how the JWT can be store in a cookie, rather than local storage and therefor be used on the server for SSR.

## How does it work?:

The express app (the server) has a `/authSuccess` route setup which will setup the client cookie with the auth token. The idea is that the OAuth provider will be given `/authSucess` as a redirect URL

When the server receives a request, it initialises the AuthContext value to `authenticated: true` if the client has a valid `nexusAuth` cookie, or `authenticated: false` is no valid cookie is present.

The client read its own cookies and then setups up the AuthContext the same way the server does.

Finally, in the Router, the `PrivateRoute` component checks if the user is authenticated from the AuthContext and redirects to the `/login` page if not.

So basically, from the app point of view, the flow is:

```
authData;
if (client) authData = setAuthFromLocalCookie()
if(server) authData = setAuthtFromRequestCookie()

renderAppWithAuthContext(authData)
```

### DIFF
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -2,18 +2,48 @@ import React = require('react');
 import ReactDOM = require('react-dom');
 import { BrowserRouter } from 'react-router-dom';
 import App from '../shared/App';
+import AuthContext, { AuthContextState } from '../shared/context/AuthContext';
 
 const rawBase: string = (window as any)['__BASE__'] || '/';
 // remove trailing slash
 const base: string = rawBase.replace(/\/$/, '');
 
+// get auth data from cookies
+// TODO: this is a POC, this code needs to be improved and tested
+function getAuthData(): AuthContextState {
+  const all: string[] = decodeURIComponent(document.cookie).split(';');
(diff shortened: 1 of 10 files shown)


mdupont:
Please also see https://infosec.mozilla.org/guidelines/web_security.html#cookies and https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet

src/client/index.tsx
```diff
@@ -2,18 +2,48 @@ import React = require('react');
...
+  <AuthContext.Provider value={getAuthData()}>
+    <BrowserRouter basename={base}>
+      <App />
+    </BrowserRouter>
+  </AuthContext.Provider>,
   document.getElementById('app'),
 );
 
+// DEVELOPMENT ONLY
+// if Hot module Replacement is enables
```
mdupont: typo

julienmachon:
Security options will be addressed in BlueBrain/nexus#228

PR #55, opened 2018-11-07
add samesite + host prefix + secure

julienmachon:
I've added:

- `SameSite` which disables cross-origin requests and set it to `strict` so we only send the cookie when site is directly navigated to
- `Secure` so cookies are only sent over HTTPS
- `__Host-` prefix which means we only send cookies over HTTPS and subdomains aren't allowed


julienmachon:
Fixes BlueBrain/nexus#253

PR #57, opened 2018-11-13
Login component

julienmachon:
Here we go, basically the same as search app:
screenshot from 2018-11-13 11-02-14

### DIFF
--- a/src/shared/components/Header/Header.stories.tsx
+++ b/src/shared/components/Header/Header.stories.tsx
@@ -1,11 +1,12 @@
-import React = require('react');
+import * as React from 'react';
 import { storiesOf } from '@storybook/react';
 import { withInfo } from '@storybook/addon-info';
 
 import Header from './index';
 
-storiesOf('Components/Header', module)
-  .add('Header', withInfo(`
+storiesOf('Components/Header', module).add(
+  'Header',
+  withInfo(`
     The header component is used throughout the application.
 
     It is used to display the user name and navigation links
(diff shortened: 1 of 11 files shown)


src/shared/components/Header/Header.stories.tsx
```diff
@@ -16,10 +17,12 @@ storiesOf('Components/Header', module)
...
-    name="Mark Hamil"
-    links={[
-      <a href="#">Home</a>,
-      <a href="#">Settings</a>,
-    ]} /></div>,
-));
+  `)(() => (
+    <div style={{ margin: '50px 40px 0px' }}>
+      <Header
+        name="Mark Hamil"
```
mdupont: Hamill

src/shared/components/Login/Login.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import Login from './index';
+
+storiesOf('Components/Login', module).add(
+  'Login',
+  withInfo(`
+    The Login component allows to redirect to the Auth Provider page.
+
+    ~~~js
+      <Login loginURL="https://keycloak.com?redirect=https://nexus.com" />
```
mdupont: nexus.com domain already used by something else

src/shared/components/Login/Login.test.tsx
```diff
@@ -0,0 +1,23 @@
+import React = require('react');
+import { shallow, mount } from 'enzyme';
+
+import Login from './index';
+
+const shallowLogin = shallow(<Login loginURL="https://keycloack.com" />);
```
mdupont: keycloak.org (.org, no c)

src/shared/components/Header/Header.stories.tsx
julienmachon: :fearful:

src/shared/components/Login/Login.stories.tsx
julienmachon: It's just an example but I'll use a something we own

PR #75, opened 2018-12-10
324 breadcrumbs

kenjinp:
# NEW
- breadcrumbs, fully configurable according to your routes file!
- passes state to breadcrumbLabel components
- breadcrumbLabels live under /breadcrumbs in views, so you can add your own special breadcrumb component to match your route!

## Minor updates
- added inline styling to header component for reducing headaches in dev mode
- refactored reducers to add orgsFetching, projectsFetching, and ResourcesFetching flags

### DIFF
--- a/src/shared/components/Breadcrumb/Breadcrumb.test.tsx
+++ b/src/shared/components/Breadcrumb/Breadcrumb.test.tsx
@@ -1,0 +1,72 @@
+import React = require('react');
+import { mount } from 'enzyme';
+import getBreadcrumbs from './getBreadcrumbs';
+import { RouteWithData } from '../../routes';
+const routes: RouteWithData[] = [
+  {
+    path: '/',
+    exact: true,
+    component: () => <div>home</div>,
+    breadcrumbLabel: () => <span>home</span>,
+  },
+  {
+    path: '/login',
+    component: () => <div>login</div>,
+  },
+  {
+    path: '/:org',
+    exact: true,
+    component: () => <div>Org</div>,
+    breadcrumbLabel: () => <span>org</span>,
+  },
+  {
+    path: '/:org/:project',
+    exact: true,
+    component: () => <div>home</div>,
+    breadcrumbLabel: () => <span>project</span>,
+  },
+];
+
+describe('Breadcrumbs', () => {
+  describe('getBreadcrumbs()', () => {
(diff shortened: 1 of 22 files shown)


src/shared/components/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,72 @@
+import React = require('react');
+import { mount } from 'enzyme';
+import getBreadcrumbs from './getBreadcrumbs';
+import { RouteWithData } from '../../routes';
+const routes: RouteWithData[] = [
```
mdupont: I have the feeling we'll end up having "real" routes vastly diverging from this set of test routes. How could we have 1 set of routes for which we override the component in order to do tests?

src/shared/components/Breadcrumb/getBreadcrumbs.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const matches: Breadcrumb[] = [];
+
+  // First need to check if there is a default path for the "home page"
+  const homeMatch = maybeMatchRouteBreadcrumbWithPath(routes, '/', state);
+  if (!!homeMatch) {
+    matches.push(homeMatch as Breadcrumb);
+  }
+
+  // Now that we've found the home breadcrumb, we can continue using this method
+  // to iterate through the pathsections as defiend as between /'s
```
mdupont: s/defiend/defined

src/shared/components/Breadcrumb/getBreadcrumbs.tsx
```diff
@@ -0,0 +1,93 @@
...
+  pathname
+    .replace(/\/$/, '')
```
mdupont: To handle the home path use case:
`.replace(/([^\/])\/$/, '$1')`

src/shared/components/Breadcrumb/Breadcrumb.test.tsx
kenjinp: We can iterate with more rigorous test cases later, right now these are simple examples that test the basic use cases

src/shared/components/Breadcrumb/getBreadcrumbs.tsx
kenjinp: doesn't pass the tests :(

PR #77, opened 2018-12-11
Fix Redux DevTools error

mdupont:
For some reason the previous approach was setting composeEnhancers() to undefined after transpiling and optimization if Redux DevTools was not available. This approach makes sure we always have a valid composeEnhancers() function.

### DIFF
--- a/src/shared/store/index.ts
+++ b/src/shared/store/index.ts
@@ -17,11 +17,11 @@ export type Services = {
 
 export type ThunkAction = ThunkAction<Promise<any>, object, Services, any>;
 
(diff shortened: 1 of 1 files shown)


src/shared/store/index.ts
```diff
@@ -17,11 +17,11 @@ export type Services = {
 
 export type ThunkAction = ThunkAction<Promise<any>, object, Services, any>;
 
-let composeEnhancers = compose;
+let composeEnhancers: Function;
```
julienmachon: Very nice catch! Very weird behaviour though...

PR #78, opened 2018-12-13
enable pagination in the project view

kenjinp:


src/shared/views/Project.tsx
```diff
@@ -21,15 +27,31 @@ const ProjectView: React.FunctionComponent<ProjectViewProps> = ({
   busy,
   resources,
   fetchResources,
+  fetching,
+  resourcePaginationSettings,
   match,
 }) => {
+  console.log({ resourcePaginationSettings });
```
mdupont: leftover console.log

julienmachon:
I don't think you've updated the `loadData` method of the route here (used for server side rendering):

https://github.com/BlueBrain/nexus-explorer/pull/78/files#diff-531be4588f505195388f950d7ccd50a4R51

Also, what's the effort of reflecting the pagination state in the URL: i.e. https://bbp-nexus.epfl.ch/staging/web/kenny/search?page=2
you get the state of your paginated list from here and you can use it both server and client side.

If that's a big job I guess it could be done in a separate task.

src/shared/routes.tsx
```diff
@@ -31,6 +32,7 @@ const routes: RouteWithData[] = [
   },
   {
     path: '/login',
+    breadcrumbLabel: LoginBreadcrumbLabel,
```
julienmachon: This still makes me sad a little. We need like a `makeView` function that we can pass to the `component` prop of a route, something like `makeView(breadcrumbComponent, bodyComponent, footerComponent)`. 

I've created an issue for that, so we don't forget: BlueBrain/nexus#334

src/shared/routes.tsx
```diff
@@ -49,7 +51,8 @@ const routes: RouteWithData[] = [
     loadData: (state, match) =>
       fetchResources(
         match && match.params && (match.params as any)['org'],
-        match && match.params && (match.params as any)['project']
+        match && match.params && (match.params as any)['project'],
+        state && state.nexus && state.nexus.resourcePaginationSettings
```
julienmachon: `loadData` is for server-side rendering, so this means the server only use the default pagination state?

kenjinp: For the moment, yeah, but when we add url parameterization (if that's the direction we go) this should work to set the correct pagination right out of the box

PR #80, opened 2018-12-18
335 simple resource view

kenjinp:
kapture 2018-12-18 at 15 24 07

# New
Ads a simple panel to check on the selected resource while browsing a list

## This is a temporary thing!!
It's important to add value ASAP, and this simple panel is a step towards a fully-featured ResourceDetailsPage component

## Bonus
Includes a simple footer

## Known Issues
long id's will bleed outside of their container


mdupont:
CSS `overflow-wrap` is your friend to avoid overflowing content.

src/shared/components/Resources/ResourceList.tsx
```diff
@@ -1,10 +1,16 @@
...
 import ResourceItem, { ResourceItemProps } from './ResourceItem';
-
 import './Resources.less';
 
+let ReactJson: any;
+if (typeof window !== 'undefined') {
+  ReactJson = require('react-json-view').default;
+}
+
+console.log(ReactJson);
```
julienmachon: Looks like this should go away

kenjinp: I keep leaving these here maybe I should lint them out... 😅

src/shared/components/Resources/ResourceList.tsx
```diff
@@ -24,24 +30,51 @@ const ResourceList: React.FunctionComponent<ResourceListProps> = ({
...
+        <Drawer
+          width={640}
+          placement="right"
+          onClose={() => setSelectedResource(null)}
+          visible={!!selectedResource}
+        >
+          {!!selectedResource && (
+            <div>
+              <h2>{selectedResource.name}</h2>
+              <ReactJson src={selectedResource.raw} />
```
julienmachon: what happens on the server when `ReactJson` has no value? does it simply not render?

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "rootDir": "./src",
+    "rootDir": "./",
```
julienmachon: :fearful: Why this change?

julienmachon: ok, that's for fetching the package.json file...

src/shared/components/Resources/ResourceList.tsx
kenjinp: I tried to configure the linter but it seems the esLint "overrides" object doesn't work for us to target only client-side files for console linting. A shame!

src/shared/components/Resources/ResourceList.tsx
```diff
@@ -24,24 +30,51 @@ const ResourceList: React.FunctionComponent<ResourceListProps> = ({
   const { from, size } = paginationSettings;
   const totalPages = Math.floor(total / size);
   const current = Math.floor((totalPages / total) * from + 1);
+  const [selectedResource, setSelectedResource] = React.useState(
+    null as Resource | null
+  );
+  console.log({ selectedResource });
```
mdupont: console.log

src/shared/components/Resources/ResourceList.tsx
```diff
@@ -24,24 +30,51 @@ const ResourceList: React.FunctionComponent<ResourceListProps> = ({
...
-      dataSource={results}
-      pagination={{
-        total,
-        current,
-        onChange: paginationChange,
-        pageSize: DEFAULT_PAGE_SIZE,
-      }}
-      renderItem={(resource: ResourceItemProps) => (
-        <ResourceItem key={resource.id} {...resource} />
+    <React.Fragment>
```
mdupont: Minor styling issue, but what do we standardize on for Fragments? Do we use the long form <React.Fragment> or the short form <> ? I like the short form personally. @julienmachon

kenjinp: hmmm actually I think it's kind of an anti-pattern to use React.Fragment, I'm using it here because I'm not sure how else to attach "out-of-element" components (like modals). 

The word looks quite ugly, so it's probably better to keep it ugly to discourage its use.

PR #81, opened 2018-12-20
SPARQL and ElasticSearch view query UI

mdupont:


src/shared/components/RawQueryView/RawQueryView.tsx
```diff
@@ -0,0 +1,85 @@
+import * as React from 'react';
+import { Form, Input, Button, Table } from 'antd';
+import { executeRawQuery } from '../../store/actions/rawQuery';
+import { RawQueryState } from '../../store/reducers/rawQuery';
+import { connect } from 'react-redux';
+import { SparqlViewQueryResponse } from '@bbp/nexus-sdk/lib/views/SparqlView';
```
julienmachon: ah! Didn't realise internal types where not accessible directly from `@bbp/nexus-sdk`. I need to sort that out

mdupont: Still needed at the moment, let's discuss on Monday

PR #83, opened 2019-01-15
Editable, Filterable Lists

kenjinp:
# NEW

Changes the project view to show a collection of resource lists. These lists can be added, deleted, and named, and filtered, with persistence maintained  in local storage. 

# TODO

Filtering on lists


src/shared/store/actions/nexus/index.ts
```diff
@@ -161,7 +302,22 @@ export const fetchResources: ActionCreator<ThunkAction> = (
     try {
       const org: Organization = await nexus.getOrganization(orgLabel);
       const project: Project = await org.getProject(projectLabel);
-      const resources: PaginatedList<Resource> = await project.listResources(
+      const elasticSearchViews: ElasticSearchView[] = await project.listElasticSearchViews();
+      const defaultElasticSearchView: ElasticSearchView = elasticSearchViews[0];
```
mdupont: Not sure the first one in the list will always be the default one. Shouldn't we add getDefaultElasticSearchView() to the sdk and discriminate based on '@id'?

kenjinp: Thanks for adding this feature to the sdK!

src/shared/components/Lists/FilterDropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <Input
+          suffix={<Icon type="filter" className="certain-category-icon" />}
+        />
+      </AutoComplete>
+      <AutoComplete
+        className="certain-category-search"
+        dropdownClassName="certain-category-search-dropdown"
+        dropdownMatchSelectWidth={false}
+        onSelect={(value, option) => {
+          // selectSchema(value as string);
```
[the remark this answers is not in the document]
kenjinp: This is a prop, to be implemented immediately after we take care of this beast

src/shared/components/Lists/ListItem.tsx
```diff
@@ -0,0 +1,149 @@
...
+        orgProjectFilterKey,
+        orgLabel,
+        projectLabel,
+        paginationSettings,
+        query
+      )
+    ),
+});
+
+export default connect(
```
julienmachon: This should not be happening at the component level, but in the view (or "container" component). Components are stateless and not aware of redux. Looks like this has been merged so needs to be added to the refactor list.

src/shared/components/Lists/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+const mapDispatchToProps = (dispatch: any, ownProps: any) => {
+  const orgProjectFilterKey = ownProps.orgLabel + ownProps.projectLabel;
+  return {
+    createList: () => dispatch(createList(orgProjectFilterKey)),
+    initialize: () =>
+      dispatch(initializeProjectList(ownProps.orgLabel, ownProps.projectLabel)),
+  };
+};
+
+export default connect(
```
julienmachon: Same comment goes for that one

src/shared/store/index.ts
```diff
@@ -32,8 +33,10 @@ try {
 export default function configureStore(
   history: History,
   nexus: Nexus,
-  preloadedState: object = {}
+  preloadedState: any = {}
```
julienmachon: This should probably be `RootState`.

PR #93, opened 2019-01-21
Deprecate all the things

julienmachon:
You can now deprecate Orgs and Projects

Display deprecated Tag:
screenshot from 2019-01-21 10-38-36

Deprecated button added to form:
screenshot from 2019-01-21 10-38-51

Confirm popup before deprecating:
screenshot from 2019-01-21 10-38-59


src/shared/views/Landing.tsx
```diff
@@ -102,6 +105,40 @@ const Landing: React.FunctionComponent<LandingProps> = ({
...
 
+  const saveAndDeprecate = (selectedOrg: Organization) => {
+    setFormBusy(true);
+
+    deprecateOrg(selectedOrg.label, selectedOrg.rev)
+      .then(
+        () => {
+          notification.success({
+            message: 'Organization deprecated',
+            duration: 2,
```
mdupont: Maybe these durations should be stored in config somewhere, with relevant constant names like NOTIFICATION_DURATION_DEFAULT = 2 and NOTIFICATION_DURATION_PERMANENT = 0

julienmachon: Agreed, but to be honest, I think the whole thing needs to be refactored. There is a lot of code duplication + the way the UI presents this to the user is quite basic. So it's going to change a lot very soon I believe

PR #97, opened 2019-01-21
WIP: align with API v1 final

mdupont:


package.json
```diff
@@ -37,7 +38,8 @@
     "react-router-dom": "^4.3.1",
     "react-spring": "^7.2.8",
     "redux": "^4.0.1",
-    "redux-thunk": "^2.3.0"
+    "redux-thunk": "^2.3.0",
+    "use-force-update": "^1.0.0"
```
julienmachon: :thinking:  hmmm...

src/shared/components/FileUpload/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </div>
+    </div>
+  );
+};
+
+const mapStateToProps = (state: RootState) => ({});
+
+const mapDispatchToProps = (dispatch: any) => ({});
+
+export default connect(
```
julienmachon: We really need to stop exposing "connected" components. Your component should only accept props. You connect the view and pass the props down. How do you render this in storybook?

src/shared/components/Lists/index.tsx
```diff
@@ -45,9 +56,54 @@ const ListsContainer: React.FunctionComponent<ListProps> = React.memo(
...
+              }}
+            />
+            <Button
+              style={{ margin: '0.5em 0' }}
+              onClick={createList}
+              icon="plus-square"
+            >
+              New Query
+            </Button>
+            <Link to={`/${orgLabel}/${projectLabel}/graph/sparql`}>
```
julienmachon: Same goes with this I guess. The button should call a function that comes from your props. Then in your view, you can make it Link somewhere if you want.

src/shared/components/Resources/ResourceForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+  render,
+  onSuccess = () => {},
+}) => {
+  const [modalVisible, setModalVisible] = React.useState(false);
+  const [formBusy, setFormBusy] = React.useState(false);
+  const saveAndCreate = async (resourceToCreate: any) => {
+    const { schemaId, payload } = resourceToCreate;
+    setFormBusy(true);
+    try {
+      const resource = await Resource.create(
```
julienmachon: Shouldn't this be a redux action?

src/shared/components/Resources/ResourceForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+        </Form.Item>
+      </Form>
+    </Spin>
+  );
+};
+
+export const ResourceFormContainer = Form.create()(ResourceForm);
+
+interface ResourceFormModalProps {
+  project: any;
```
julienmachon: :grimacing:

src/shared/store/actions/nexus/projects.ts
```diff
@@ -53,7 +56,39 @@ export const fetchAndAssignProject: ActionCreator<ThunkAction> = (
     dispatch(fetchProjectsAction());
     try {
       const project: Project = await Project.get(orgLabel, projectLabel);
-      return dispatch(fetchProjectsFulfilledAction(project));
+      const defaultElasticSearchView: ElasticSearchView = await project.getElasticSearchView();
+      const aggregationQuery = {
+        aggs: {
+          schemas: {
+            terms: {
+              size: 999,
```
julienmachon: why 999? could it be a constant?

src/shared/components/FileUpload/index.tsx
kenjinp: Ah that's a good idea

package.json
julienmachon: we need to be careful about dependencies. The more we have, the harder it is to update them. This has 5 commits, 3 stars and wasn't updated in months. It's also about 5 lines of code... I've created an issue so we don't miss it https://github.com/BlueBrain/nexus/issues/366

PR #103, opened 2019-01-24
make sure we list all project and orgs 

julienmachon:
+ remove resourceNumber from project card


src/shared/store/actions/nexus/activeOrg.ts
```diff
@@ -60,8 +60,8 @@ export const fetchOrg: ActionCreator<ThunkAction> = orgName => {
   > => {
     dispatch(fetchOrgAction());
     try {
-      const org: Organization = await nexus.getOrganization(orgName);
-      const projects: Project[] = await org.listProjects();
+      const org: Organization = await Organization.get(orgName);
+      const projects: Project[] = await Project.list(orgName, { size: 100 });
```
julienmachon: I know it's a bit of a hack but until we implement pagination....

PR #108, opened 2019-01-28
Better button label

mdupont:


src/shared/store/actions/project.ts
```diff
@@ -219,12 +219,18 @@ export const makeProjectPublic: ActionCreator<ThunkAction> = (
       // @ts-ignore
       const endpoint = getState().config.apiEndpoint;
 
+      const publicPermissions = [
```
julienmachon: Do you need to rebase maybe?

mdupont: Yep, done :-)

PR #110, opened 2019-01-31
Minor resource accessbility updates

kenjinp:
# Minor Resource List Accessibility Changes
kapture 2019-01-31 at 15 47 40

You can now tab down to the text query interface directly from the Browser URL bar,
tab between items in the resource lists, 
press enter while resource items are in focus to bring up their details view, 
press escape to go back to the list page.


julienmachon:
I mean, fix your tests and then :fire: :fire_engine: :woman_firefighter:

kenjinp:
jenkins' fault :(

PR #111, opened 2019-02-04
Animated paginated list component

kenjinp:
# Animations placeholder & WIP


julienmachon:
Pagination for Project still missing but pagination for org is in place. The code is becoming extremely messy: spaghetti components with too many options, code duplication, naming inconsistency, not tests, etc... We need to refactor all of that next sprint.

julienmachon:
Could I get an approval if everything's fine? :kissing_heart:

PR #112, opened 2019-02-05
Resource Views

kenjinp:
# NEW
Add Modal Switch Routes - if you navigate using react router with a `state: { modal: true }` flag then it will render that route inside a modal instead, and preserver the current route rendered behind it. Useful for creating pinterest routes.

ResourceView - instead of using a Drawer component, it render's the resource inside a modal as a View with it's own route.

You can navigate to resources directly be visiting
`/:orgLabel/:projectLabel/:resourceID`

## Resource Editing
A simple form based on codemirror, with some fancy validation feedback.

new edit form

## Resource Creation
again based on codemirror.

## Resource MetaData Card
useful to display machine-generated information about a resource
screenshot 2019-02-14 at 15 24 21

## Improved @Type Tags
screenshot 2019-02-14 at 15 24 29

## Hovering Side-Panel
hovering side bar

## UserAvatar 
screenshot 2019-02-14 at 15 24 15

## Animated Route Changes
because why not


julienmachon:
Noooooo, you're doing it again!!! :grimacing:
 
That's a loooot of changes already and I can see a 3 :star: non-maintained github dependency in there... :cry:

src/shared/components/Lists/ListItem.tsx
```diff
@@ -9,6 +9,7 @@ import ResourceList from '../Resources/ResourceList';
 import { updateList, deleteList, cloneList } from '../../store/actions/lists';
 import { queryResources, makeESQuery } from '../../store/actions/queryResource';
 import ListControlPanel from './ListControlPanel';
+import { push } from 'connected-react-router';
```
julienmachon: nooooooooooo :hankey:

src/shared/components/Resources/MetaData.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+import { Card } from 'antd';
+import { Meta } from 'antd/lib/list/Item';
+import moment = require('moment');
```
julienmachon: did you try?

```typescript
import * as moment from 'moment';
```

src/shared/components/Resources/ResourceList.tsx
```diff
@@ -3,6 +3,7 @@ import { List, Drawer } from 'antd';
 import { PaginatedList, Resource, PaginationSettings } from '@bbp/nexus-sdk';
 import ResourceItem, { ResourceItemProps } from './ResourceItem';
 import './Resources.less';
+import { Link } from 'react-router-dom';
```
julienmachon: :hankey:

kenjinp:
This PR was too chunky so I'm gonna deliver it in smaller pieces like I should have in the beginning 🤦‍♀️

PR #121, opened 2019-02-14
add side menu to workspace

kenjinp:
# Side Menu

> ⚠️ this is the first part of a series of PRs to work through my bloated PR here: https://github.com/BlueBrain/nexus-web/pull/112

Adds a fixed, hovering `SideMenu` component to the workspace

hovering side bar

## Why?
because our lists may be long, and the `Drawer` component makes overflow hidden. Plus, it looks nice.

## Other changes

- updated react-spring to the latest major version, which accepts only it's custom hooks, therefore I've taken the chance to update the `AnimatedList` component and provide a storybook for it. 

-  upgrade to the latest version of React because of the new react-spring version.

-  deleted the FlipCard components as they're not used (although, could be fun in the future)


julienmachon:
This is :eggplant: !

PR #124, opened 2019-02-15
👷‍♀️FIX:  pagination settings on org list

kenjinp:
Fixes pagination settings by passing props to the fetchOrgs action

### DIFF
--- a/src/shared/store/actions/nexus/orgs.ts
+++ b/src/shared/store/actions/nexus/orgs.ts
@@ -7,6 +7,8 @@ import {
 } from '@bbp/nexus-sdk';
 import { ThunkAction } from '../..';
 import { FetchAction, FetchFulfilledAction, FetchFailedAction } from '../utils';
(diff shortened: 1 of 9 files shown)


src/shared/store/actions/nexus/orgs.ts
```diff
@@ -65,8 +65,9 @@ export const fetchOrgs: ActionCreator<ThunkAction> = (
   > => {
     dispatch(fetchOrgsAction());
     try {
+      console.log({ paginationSettings });
       const orgs: PaginatedList<Organization> = await Organization.list({
-        size: (paginationSettings && paginationSettings.size) || 20,
+        size: (paginationSettings && paginationSettings.size) || 5,
```
julienmachon: Use variable here maybe?

kenjinp: I'll add a reducer

julienmachon:
Whoop Whoop!

PR #127, opened 2019-02-15
Fancy Type Tags

kenjinp:
Just some nicer `@type` tags, implemented using an icon hash and color-thief.

screenshot 2019-02-15 at 17 26 33

### DIFF
--- a/src/shared/components/Types/TypesIcon.tsx
+++ b/src/shared/components/Types/TypesIcon.tsx
@@ -1,8 +1,19 @@
 import * as React from 'react';
-import { Avatar, Tooltip } from 'antd';
+import { Avatar, Tooltip, Tag } from 'antd';
 import * as Identicon from 'identicon.js';
 import './Types.less';
-import * as md5 from "md5"
+import * as md5 from 'md5';
+const { getPaletteFromURL } = require('color-thief-node');
+
+// https://stackoverflow.com/questions/5623838/rgb-to-hex-and-hex-to-rgb
+function componentToHex(component: number) {
+  const hex = component.toString(16);
+  return hex.length === 1 ? `0${hex}` : hex;
+}
+
+function rgbToHex(r: number, g: number, b: number) {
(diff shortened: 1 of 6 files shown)


src/shared/components/Types/TypesIcon.tsx
```diff
@@ -1,8 +1,19 @@
 import * as React from 'react';
-import { Avatar, Tooltip } from 'antd';
+import { Avatar, Tooltip, Tag } from 'antd';
 import * as Identicon from 'identicon.js';
 import './Types.less';
-import * as md5 from "md5"
+import * as md5 from 'md5';
+const { getPaletteFromURL } = require('color-thief-node');
```
julienmachon: can you not do an `import { getPaletteFromURL } from 'color-thief-node` ?

src/shared/components/Types/TypesIcon.tsx
```diff
@@ -13,20 +24,36 @@ const TypesIcon: React.SFC<TypesIconProps> = ({ type }) => {
   return (
     <ul className="types-list">
       {type.map((type, index) => {
+        const [colors, setColor] = React.useState<string[] | null>(null);
```
julienmachon: I am not sure that you can use React hooks in a map like that, plus it's in the return statement of your component which I don't think is okay either.

You need to make sure you follow that https://reactjs.org/docs/hooks-rules.html

kenjinp: Good to know! fixed it!

src/shared/components/Types/TypesIcon.tsx
kenjinp: I think I was being lazy, if you use import then typescript will complain that this dep doesn't have a declaration file. I've added a type dec in our `custom.d.ts` file to fix that.

It's a wacky dependency but I think it's better than the original library because of node support, which I have plans for...

PR #129, opened 2019-02-18
return error in action

julienmachon:
That should fix the nasty error


julienmachon:
We need to fix the typing of our Redux Action. The issue has been addressed in BlueBrain/Nexus#452

PR #130, opened 2019-02-18
Resource Details Page

kenjinp:
# Resource Pages!
> as part of "project dismantle https://github.com/BlueBrain/nexus-web/pull/112"

## NEW
- viewing/editing resources
- resource details page (somewhat sparse)
- links from the resource list

## Caveats
- It's missing the metaData, which will be added in a subsequent PR. 
- Because the pinterest route is hard to implement, the resource view will not appear in a modal until I make another PR for that. 
- You cannot edit things like views or schemas. I'm not sure why, I think the API has changed? This must be fixed at the SDK level

### DIFF
--- a/src/shared/components/Resources/ResourceDetails.less
+++ b/src/shared/components/Resources/ResourceDetails.less
@@ -1,0 +1,2 @@
+.resource-details {
+}
--- a/src/shared/components/Resources/ResourceDetails.less
+++ b/src/shared/components/Resources/ResourceDetails.less
@@ -1,2 +1,0 @@
-.resource-details {
-}
--- a/src/shared/components/Resources/ResourceDetails.tsx
+++ b/src/shared/components/Resources/ResourceDetails.tsx
@@ -1,0 +1,66 @@
+import * as React from 'react';
+import { Resource } from '@bbp/nexus-sdk';
+import { Spin, Button, notification, Icon } from 'antd';
+import './ResourceDetails.less';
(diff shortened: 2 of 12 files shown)


src/shared/components/Resources/ResourceDetails.less
```diff
@@ -0,0 +1,2 @@
+.resource-details {
```
mdupont: ???

src/shared/components/Resources/ResourceDetails.tsx
```diff
@@ -0,0 +1,66 @@
...
+const ResourceDetails: React.FunctionComponent<ResourceViewProps> = props => {
+  const { resource, error, isFetching, onSuccess } = props;
+  const [editing, setEditing] = React.useState(false);
+  const [busy, setFormBusy] = React.useState(isFetching);
+
+  const handleSubmit = async (value: any) => {
+    if (resource) {
+      try {
+        setFormBusy(true);
+        console.log(resource);
```
mdupont: 🔫

src/shared/components/Resources/ResourceDetails.less
kenjinp: whoops

src/shared/components/Resources/ResourceDetails.tsx
kenjinp: 💥

PR #131, opened 2019-02-18
Revert all the things

julienmachon:
This reverts nexus-web to a buildable state.
We don't have to merge it if we manage to fix the current build. It is here just in case


julienmachon:
No Need, kenny has fixed all the things. :pray:

PR #133, opened 2019-02-18
Uploading Files

kenjinp:
# Uploading Files

Implements the `NexusFile.create()` method from the SDK as a nice little drag-drop file uploader widget. 

upload files


src/shared/components/FileUpload/index.tsx
```diff
@@ -1,22 +1,43 @@
 import * as React from 'react';
 import { Upload, Icon, message, Switch } from 'antd';
-import { connect } from 'react-redux';
```
julienmachon: :pray: :bee:

PR #135, opened 2019-02-19
Meta data card

kenjinp:
#Meta Data Card
- displays useful info
- includes new types
- User Avatar (not so pretty yet)
- various error handling and bug fixes
screenshot 2019-02-19 at 12 57 30
screenshot 2019-02-19 at 12 57 34

### DIFF
--- a/src/shared/components/Footer/Footer.less
+++ b/src/shared/components/Footer/Footer.less
@@ -7,7 +7,7 @@
   width: 100%;
   bottom: 0;
   margin-top: 2em;
-  position: absolute;
+  // position: absolute;
   left: 0;
   padding: 0.5em;
   > .footerlet {
--- a/src/shared/components/Footer/Footer.less
+++ b/src/shared/components/Footer/Footer.less
@@ -7,7 +7,6 @@
   width: 100%;
   bottom: 0;
   margin-top: 2em;
-  // position: absolute;
   left: 0;
   padding: 0.5em;
   > .footerlet {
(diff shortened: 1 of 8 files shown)


src/shared/components/Footer/Footer.less
```diff
@@ -7,7 +7,7 @@
   width: 100%;
   bottom: 0;
   margin-top: 2em;
-  position: absolute;
+  // position: absolute;
```
mdupont: boo

src/shared/components/Resources/ResourceDetails.tsx
```diff
@@ -19,10 +20,7 @@ const ResourceDetails: React.FunctionComponent<ResourceViewProps> = props => {
     if (resource) {
       try {
         setFormBusy(true);
-        await resource.update({
-          context: resource.context,
-          ...value,
-        });
+        await resource.update(value);
```
julienmachon: This should be dispatching an action, rather than calling the SDK directly

src/shared/components/Resources/ResourceDetails.tsx
```diff
@@ -41,20 +39,33 @@ const ResourceDetails: React.FunctionComponent<ResourceViewProps> = props => {
       }
     }
   };
+  console.log(resource && { raw: resource.raw });
```
julienmachon: We really need to add that linter rule in place but in the meantime, I find it useful to go through my own file changes. These things are kind of easy to spot

src/shared/components/User/Avatar.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { Tooltip, Avatar } from 'antd';
+
+export interface AvatarProps {
+  createdBy: string;
+}
+
+const UserAvatar: React.FunctionComponent<AvatarProps> = props => {
+  const { createdBy } = props;
+  const [name] = createdBy.split('/').slice(-1);
```
julienmachon: This is confusing. `CreatedBy` is a string, how do we know we have a / in it?

PR #142, opened 2019-02-20
Replace bread crumb

julienmachon:
The goal is to replace the orgs and project landing pages that nobody likes (based on feedback, including ours).

The plan is to have an Org and Project selector in the header instead that will allow to both navigate and select items. The good news is that I make a Org selector Component when I started Nexus Web, so we just need to adapt it.

There are a few steps to get there but this PR includes:
- remove that breadcrumb as it is buggy and takes a lot of space in terms of code and API calls
- make space for the org/project selectors by moving the logo in the middle (like Trello a little)
- update org selector component style

Next, we need to:
- create the Project Selector component BlueBrain/nexus#458
- add both selectors to the Header and hook then to actions/state BlueBrain/nexus#459
- design new landing page BlueBrain/nexus#460
- remove old landing pages BlueBrain/nexus#461

This is how it looks like:
screenshot from 2019-02-20 11-50-19

screenshot from 2019-02-20 13-57-29
screenshot from 2019-02-20 13-57-44
screenshot from 2019-02-20 13-57-53


kenjinp:
yay!

PR #146, opened 2019-02-25
Image Previews

kenjinp:
# image preview support in lists

- Adds a component to demonstrate `NexusFile` class

screenshot 2019-02-25 at 16 25 16

## Caveat 
I'm thinking we'll need to move the PaginatedList Resource responses, for example from the ElasticSearchView response, to create the proper classes instead of just Resource, so we can better use typing and instances to switch between specialised components / views.

### DIFF
--- a/src/shared/components/Resources/ResourceItem.tsx
+++ b/src/shared/components/Resources/ResourceItem.tsx
@@ -5,7 +5,7 @@ import TypesIcon from '../Types/TypesIcon';
 import './Resources.less';
 import ResourceMetadataCard from './MetadataCard';
 import { Resource } from '@bbp/nexus-sdk';
-import ResourcePreview from './ResourcePreview';
+import ResourcePreview, { hasDisplayableImage } from './ResourcePreview';
 
 const MOUSE_ENTER_DELAY = 0.5;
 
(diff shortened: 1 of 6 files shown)


src/shared/components/Resources/ResourceItem.tsx
```diff
@@ -19,6 +20,16 @@ const ResourceListItem: React.FunctionComponent<ResourceItemProps> = props => {
   const { resource, index, onClick = () => {} } = props;
   const containerRef = React.createRef<HTMLDivElement>();
 
+  let Preview = null;
+  if (
```
mdupont: should be a separate function hasDisplayableImage() or something similar

src/shared/components/Resources/ResourcePreview.tsx
```diff
@@ -0,0 +1,48 @@
...
+      resource.self,
+      resource.orgLabel,
+      resource.projectLabel,
+      true
+    )
+      .then((nexusFile: NexusFile) => {
+        setFile(nexusFile);
+      })
+      .catch((error: Error) => {
+        // Handle error?
```
mdupont: at least log it

src/shared/components/Resources/ResourcePreview.tsx
```diff
@@ -0,0 +1,48 @@
...
+      });
+  }
+
+  return (
+    <Popover content={'hello'} mouseEnterDelay={MOUSE_ENTER_DELAY}>
+      {file ? (
+        <Avatar
+          shape="square"
+          src={`data:${file.mediaType};base64,${file.rawFile as string}`}
```
mdupont: as we sure it will always be base64 encoded?

kenjinp: yes, the sdk only supports base64 for the moment until other file support is gradually added. And this is an image previewer, so we'll probably only use a base64 output anyways.

src/shared/components/Resources/ResourceItem.tsx
kenjinp: good idea :)

PR #151, opened 2019-02-27
Forced Graph Dot Visualizer

kenjinp:
# Graph Vis.
Displays a physics simulated graph

screenshot 2019-02-27 at 18 15 29

## TODO: 
- [x] Node labels
- [x] Edge labels
- [x] dot-to-Node/Links parser
- [ ] links to resources? (I'll take it up in next iteration)


src/shared/components/Graph/GraphVisualizer.less
```diff
@@ -0,0 +1,7 @@
+@import '../../lib.less';
+
+.graph-visualizer {
+  .interactiveArea();
+  height: auto;
+  min-height: 600px;
```
mdupont: why this number?

kenjinp: The hight must be defined for this svg stuff, and I just think 600 looks pretty nice. So it's completely arbitrary

kenjinp:
Some progress

There was a lot to do about making sure the context was not rendering and that state wasn't stale. I think we have similar problems in other places around the application, so that's neat that I was able to squash this bug. It caused some infinite loading problems when using React.useState.

Next up is labels for links / edges.

screenshot 2019-02-28 at 17 58 07

src/shared/components/Graph/GraphVisualizer.less
mdupont: Better indicate it explicitely in the code then :-)

src/shared/components/Resources/ResourceDetails.tsx
```diff
@@ -1,21 +1,27 @@
...
+  dotGraph: string | null;
 }
 
 const ResourceDetails: React.FunctionComponent<ResourceViewProps> = props => {
-  const { resource, error, isFetching, onSuccess } = props;
+  const { resource, error, isFetching, onSuccess, dotGraph } = props;
   const [editing, setEditing] = React.useState(false);
   const [busy, setFormBusy] = React.useState(isFetching);
 
+  const isFile = resource && resource.type && resource.type.includes('File');
```
mdupont: 'File' should be a constant

src/shared/components/Graph/Graph.ts
```diff
@@ -0,0 +1,39 @@
...
+  constructor(dotGraph: string, element: HTMLElement) {
+    this.dotGraph = dotGraph;
+    this.element = element;
+  }
+
+  render() {
+    try {
+      this.destroy = graph(this.dotGraph, this.element);
+    } catch (error) {
+      // tslint:disable-next-line:no-console
```
julienmachon: Be careful with this, it might end up in the code base. You can skip the githook when you push instead. That way we can insure no dead code makes it through

kenjinp: That's cool, but this particular error I want to log, in case we get some interesting data that breaks the parser.

kenjinp: Ok, not in prod, we'll use another consolidated logging service somehow 👍

kenjinp:
Labels!

Caveats for this version
- No arrows pointing to target, which would be nice
- if a node shares two edges with another, there will be label overlap

dotgraph

src/shared/views/Resource.tsx
```diff
@@ -20,28 +22,53 @@ interface ResourceViewProps {
...
+  }, [match.params.resourceId]);
+
+  // Once we have a legit resource, let's fetch the DOT
+  React.useEffect(() => {
+    if (resource) {
+      setBusy(true);
+      // cannot use instance method here,
+      // because sometimes we'll get 'resource'
+      // as a raw js object from the server and not an instance
+      Resource.getSelfRawAs(resource.self, ResourceGetFormat.DOT)
```
julienmachon: This should probably be a redux action

julienmachon: also, when was this method added? It isn't documented https://github.com/BlueBrain/nexus-sdk-js#resources-1

julienmachon:
Looks awesome!

src/shared/views/Resource.tsx
kenjinp: updated readme: https://github.com/BlueBrain/nexus-sdk-js/pull/94

PR #153, opened 2019-03-06
Advanced settings section for Project from

mdupont:
Hide base, vocab and api mappings fields under an "Advanced settings" collapsed panel so that users know not to touch them unless they know what they are doing.

fixes BlueBrain/nexus#486


src/shared/components/Projects/__tests__/ProjectForm.spec.tsx
```diff
@@ -1,12 +1,21 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
+
+// Need to mock Date.now() because it is used down the line
+// to generate a per-element class, which breaks snapshot testing.
+// Must be done before importing the React component.
+jest.spyOn(Date, 'now').mockImplementation(() => 1234567890123);
```
julienmachon: :boom:

PR #154, opened 2019-03-06
444 server side error codes

kenjinp:
# New! Errors & Codes! 🚨

```typescript
<Status code={!!error ? error.code : 200}>
    <ResourceView
        dotGraph={dotGraph}
         ...
      />
 </Status>
```

- new `RequestErrors` types and classes. They have `codes`
- fetching resources using the `createFetchReducer` will expect` RequestErrors` to be thrown on failure, like when the SDK throws an error.
- new `Status` component that takes a `code` and passes it to the server when rendering using the router context
- all this results in reliable error status codes being rendered by the application

associated issue: https://github.com/BlueBrain/nexus/issues/444


src/shared/store/actions/utils/errors.ts
```diff
@@ -0,0 +1,33 @@
+export class Forbidden extends Error {
+  readonly code: number = 403;
```
mdupont: Same remark about numbers, they should be constants with names, like HTTP_STATUS_UNAUTHORIZED

src/shared/store/actions/utils/errors.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+export class NotFound extends Error {
```
mdupont: Why create 1 class per error type instead of one global "RequestError" class that had a "code" property with the status code? Wouldn't it be better to have types based on a definition like: "type Forbidden is an instance of RequestError class, only if that instance has a status code of HTTP_STATUS_UNAUTHORIZED". Sorry I haven't the exact Typescript snippet here.

src/shared/store/actions/utils/errors.ts
```diff
@@ -0,0 +1,33 @@
...
+  NotFound,
+  ServiceUnavailable,
+};
+
+export const formatError = (error: Error) => {
+  const errorType = error.message.replace(/ /g, '');
+  if (requestErrorTypes[errorType]) {
+    return new requestErrorTypes[errorType](error.message);
+  }
+  return new BadRequest(error.message);
```
mdupont: This would return an erroneous code if the individual status code is not explicitely handled in the list below. This problem can be solved if using my proposal of making RequestError a class and individual error types, well, types

src/shared/views/Project.tsx
```diff
@@ -41,87 +43,86 @@ const ProjectView: React.FunctionComponent<ProjectViewProps> = ({
...
-          <Empty
-            style={{ marginTop: '22vh' }}
-            description="No project data found here..."
-          />
-        </>
-      )}
-      {project && (
-        <>
-          <div>
+    <Status code={!!error ? error.code : 200}>
```
mdupont: Same remark about constants instead of magic numbers

src/shared/views/Project.tsx
```diff
@@ -41,87 +43,86 @@ const ProjectView: React.FunctionComponent<ProjectViewProps> = ({
...
+      <div className="project">
+        {!project && error && (
+          <>
+            <h1 style={{ marginBottom: 0, marginRight: 8 }}>
+              {match.params.project}
+            </h1>
+            <Empty
+              style={{ marginTop: '22vh' }}
+              description="There was a problem while loading this project!"
```
mdupont: Shouldn't we add the error message text along with this text?

src/shared/views/Resource.tsx
```diff
@@ -36,13 +38,15 @@ const ResourceViewPage: React.FunctionComponent<ResourceViewProps> = props => {
...
     <>
       {!!resource && <Helmet title={resource.name} />}
-      <ResourceView
-        dotGraph={dotGraph}
-        onSuccess={fetch}
-        resource={resource}
-        error={error}
-        isFetching={isFetching}
-      />
+      <Status code={!!error ? error.code : 200}>
```
mdupont: magic number

PR #155, opened 2019-03-07
469 project creation delay

kenjinp:
# Improved Project Creation Flow ⛲️

> Solving issue: https://github.com/BlueBrain/nexus/issues/469

On project creation, the app will try to query that project's default ES view every .5s. If there's content there, the project is ready and the user will be redirected .

projectscaffolding

This polling process will last 30 seconds. If still we didn't find that the project is ready, we will redirect to the project landing page anyways with a nice warning that thing's aren't quite read yet, but won't freeze any user actions. 

projectistakingtoolong

### DIFF
--- a/src/shared/store/actions/project.ts
+++ b/src/shared/store/actions/project.ts
@@ -5,6 +5,7 @@ import { CreateProjectPayload } from '@bbp/nexus-sdk/lib/Project/types';
 import { httpGet, httpPut } from '@bbp/nexus-sdk/lib/utils/http';
 import { IdentityResponse, Identity } from '@bbp/nexus-sdk/lib/ACL/types';
 import { asyncTimeout } from '../../utils';
+import { notification } from 'antd';
 
 const DEFAULT_ELASTIC_SEARCH_INDEX_ID = 'nxv:defaultElasticSearchIndex';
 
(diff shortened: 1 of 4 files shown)


src/shared/store/actions/project.ts
```diff
@@ -1,9 +1,13 @@
...
-import { Project } from '@bbp/nexus-sdk';
+import { Project, ElasticSearchView } from '@bbp/nexus-sdk';
 import { ThunkAction } from '..';
 import { CreateProjectPayload } from '@bbp/nexus-sdk/lib/Project/types';
 import { httpGet, httpPut } from '@bbp/nexus-sdk/lib/utils/http';
 import { IdentityResponse, Identity } from '@bbp/nexus-sdk/lib/ACL/types';
+import { asyncTimeout } from '../../utils';
+import { notification } from 'antd';
+
+const DEFAULT_ELASTIC_SEARCH_INDEX_ID = 'nxv:defaultElasticSearchIndex';
```
mdupont: Don't we have this elsewhere already? Or as part of the SDK?

mdupont:
Fixes BlueBrain/nexus#469

src/shared/store/actions/project.ts
kenjinp: You're right, it's actually not necessary to specify an ID if you just want the default view. Neato!

PR #179, opened 2019-03-21
555 list available realms

julienmachon:
That's it, we no get realms from nexus and let the user choose which one they want to authenticate with in the event of multiple realms available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,15 +54,15 @@ docker build . --tag=nexus-web
 
 - `BASE_PATH`: The base of the app: i.e. `/staging/web` if hosted on `https://bbp-nexus.epfl.ch/staging/web` (default is `/`)
 - `HOST_NAME`: name of host where application is available from: i.e. `https://bbp-nexus.epfl.ch` (default is protocol + host where server is running from)
-- `CLIENT_ID`: The application name used for _OpenID Connect_ authentication
+- `CLIENT_ID`: The application name used for _OpenID Connect_ authentication (default is `nexus-web`)
 - `API_ENDPOINT`: The URL pointing to Nexus API. Default is '/'
-- `AUTH_ENDPOINT`: _OpenID Connect_ Authentication endpoint. Empty by default (no auth).
-- `LOGOUT_ENDPOINT`: _OpenID Connect_ Logout endpoint. Empty by default (no logout).
+- `SECURE`: Is nexus web running in https or not. Default is `true`
 
 ## Getting involved
 
 Issue tracking is centralized into [the main Blue Brain Nexus repository](https://github.com/BlueBrain/nexus).
 
 There are several channels provided to address different issues:
+
 - **Feature request**: If there is a feature you would like to see in this application, please first consult the [list of open feature requests](https://github.com/BlueBrain/nexus/issues?q=is%3Aopen+is%3Aissue+label%3Afeature+label%3Afrontend+label%3Anexus-web). In case there isn't already one, please [open a feature request](https://github.com/BlueBrain/nexus/issues/new?labels=feature,frontend,nexus-web) describing your feature with as much detail as possible.
 - **Bug report**: If you have found a bug, please create an issue [here](https://github.com/BlueBrain/nexus/issues/new?labels=bug,frontend,nexus-web).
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -54,15 +54,15 @@ docker build . --tag=nexus-web
 
 - `BASE_PATH`: The base of the app: i.e. `/staging/web` if hosted on `https://bbp-nexus.epfl.ch/staging/web` (default is `/`)
 - `HOST_NAME`: name of host where application is available from: i.e. `https://bbp-nexus.epfl.ch` (default is protocol + host where server is running from)
-- `CLIENT_ID`: The application name used for _OpenID Connect_ authentication
+- `CLIENT_ID`: The application name used for _OpenID Connect_ authentication (default is `nexus-web`)
 - `API_ENDPOINT`: The URL pointing to Nexus API. Default is '/'
-- `AUTH_ENDPOINT`: _OpenID Connect_ Authentication endpoint. Empty by default (no auth).
-- `LOGOUT_ENDPOINT`: _OpenID Connect_ Logout endpoint. Empty by default (no logout).
+- `SECURE`: Is nexus web running in https or not. Default is `true`
```
mdupont: SECURE should not be assumed to be true if we don't know the context, it should always be based on the URL scheme currently in use (detection on the server side) and fall back to false

src/shared/utils/index.ts
```diff
@@ -138,3 +139,31 @@ export const labelOf = (string: string) => {
...
+ * @param identities
+ * @param realms
+ */
+export function getLogoutUrl(
+  identities: Identity[],
+  realms: { label: string; endSessionEndpoint: string }[]
+): string {
+  // find authenticated Identity and get realm name
+  const auths = identities.filter(
+    identity => identity['@type'] === 'Authenticated'
```
mdupont: "Authenticated" is a magic string, should be a constant or a config item, imported from a central storage place for these kinds of values

src/shared/utils/index.ts
```diff
@@ -138,3 +139,31 @@ export const labelOf = (string: string) => {
...
+ *
+ * @param identities
+ * @param realms
+ */
+export function getLogoutUrl(
+  identities: Identity[],
+  realms: { label: string; endSessionEndpoint: string }[]
+): string {
+  // find authenticated Identity and get realm name
+  const auths = identities.filter(
```
mdupont: This operation could be a `.find()` returning either the matching identity or a realm, since you only care about the 1st match.

src/shared/views/Login.tsx
```diff
@@ -2,20 +2,24 @@ import * as React from 'react';
...
   redirectUrl: string;
+  redirect(): void;
 }
 
 const Login: React.FunctionComponent<LoginViewProps> = props => {
-  const realms: Realm[] = [
-    { name: 'BBP', authorizationEndpoint: props.authorizationEndpoint },
-  ];
+  const { realms, redirect } = props;
+  if (realms.length === 0 || !realms) {
```
mdupont: the second test statement would also work in the first case, can be simplified to `!realms`

src/shared/utils/index.ts
julienmachon: It's not, it's a UNION type. You get a compile error if you don't use the right values.

src/shared/utils/index.ts
julienmachon: yes! that's the one I was looking for. For some reason I keep confusing it with `.some()` which only returns a boolean

src/shared/utils/index.ts
julienmachon: https://github.com/BlueBrain/nexus-sdk-js/blob/master/src/ACL/types.ts#L3

README.md
julienmachon: It cannot be based on the URL as, in our case for instance, the service is running in OpenShift, behind a reverse proxy. This flag is to is to fix that very issue we had in our dev environment for example.

src/shared/utils/index.ts
julienmachon: fixed in 7172506e898cf880598ca2dfa8d90588445f222f

src/shared/views/Login.tsx
julienmachon: fixed in 7172506e898cf880598ca2dfa8d90588445f222f

PR #180, opened 2019-03-21
chore: update deps and fix routing issue

mdupont:


src/server/index.tsx
```diff
@@ -33,16 +33,20 @@ const app: express.Express = express();
...
 const base: string = rawBase.replace(/\/$/, '');
+// Utility fonction to compute the full path depending on the base path
+const buildPath = buildPathFactory(base);
+// base path ending with a slash, for routing purpose
+const fullBasePath = buildPath();
 // enable logs
 app.use(morgan('dev'));
 // expose status route
-app.get(`${base}/status`, (req, res) => res.send('OK'));
+app.get(buildPath('status'), (req, res) => res.send('OK'));
```
julienmachon: Hmm, not sure... We're actually using path the wrong way. We should create normal express routers and mount it at `${base}` if there's one.

Also, we need to re-configure the service at openshift level

mdupont:
Outdated, now fixed. Will open a separate PR for deps update

PR #185, opened 2019-03-27
✨FEAT: List Item component

kenjinp:
An everyday hard working list element component
- should go on to replace orgs, projects and resource listings
- has mandatory `label` and optional `description`, `details`, and `action` sections
- included some react hook lint support
Screenshot 2019-03-27 at 17 28 44

### DIFF
--- a/src/shared/components/Animations/ListItem.stories.tsx
+++ b/src/shared/components/Animations/ListItem.stories.tsx
@@ -1,0 +1,84 @@
+import * as React from 'react';
+import { storiesOf } from '@storybook/react';
+import { withInfo } from '@storybook/addon-info';
+import {
+  withKnobs,
+  text,
+  array,
+  object,
+  boolean,
+} from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
+import ListItem from './ListItem';
+import { Tag, Button } from 'antd';
+
+const exampleItems = [
+  {
+    label: 'nexus',
+    projectNumber: 1200,
+    description:
+      "Yeah, but your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should. You're a very talented young man, with your own clever thoughts and ideas. Do you need a manager? They're using our own satellites against us. And the clock is ticking.",
+  },
+  { label: 'bbp', projectNumber: 300, deprecated: true, action: true },
+  { label: 'hbp', projectNumber: 1 },
+  {
+    label: 'nasa',
+    projectNumber: 912839,
+    description:
+      "God creates dinosaurs. God destroys dinosaurs. God creates Man. Man destroys God. Man creates Dinosaurs. Hey, you know how I'm, like, always trying to save the planet? Here's my chance. You really think you can fly that thing? Did he just throw my cat out of the window?",
+  },
+  { label: 'tesla', projectNumber: 3, deprecated: true },
+  { label: 'rolex', projectNumber: 3424 },
+];
+
+storiesOf('Components/ListItem', module)
+  .addDecorator(withKnobs)
+  .add(
+    'ListItem',
+    withInfo(`
+    🗒 Try this on for size when you want to render consistent list elements
+
+    ~~~js
+    <ListItem label={"Hi mom!"} />
+    ~~~
+  `)(() => {
+      return (
+        <div style={{ margin: '50px 40px 0px', width: '50%' }}>
+          <p>Good for many use cases, eh?</p>
+          <ul>
+            {exampleItems.map(element => {
+              const {
+                label,
+                action: elementAction,
+                description,
+                deprecated,
+              } = element;
+              return (
+                <ListItem
+                  label={element.label}
+                  onClick={action('element-clicked')}
+                  detail={
+                    deprecated ? <Tag color="red">deprecated</Tag> : undefined
+                  }
+                  description={description}
+                  action={
+                    elementAction ? (
+                      <Button
+                        icon="edit"
+                        className="edit-button"
+                        type="primary"
+                        tabIndex={1}
+                        onClick={() => action('edit-clicked')}
+                      />
+                    ) : (
+                      undefined
+                    )
+                  }
+                />
+              );
+            })}
+          </ul>
+        </div>
+      );
+    })
+  );
--- a/src/shared/components/Animations/ListItem.stories.tsx
+++ b/src/shared/components/Animations/ListItem.stories.tsx
@@ -46,19 +46,29 @@ storiesOf('Components/ListItem', module)
         <div style={{ margin: '50px 40px 0px', width: '50%' }}>
           <p>Good for many use cases, eh?</p>
           <ul>
-            {exampleItems.map(element => {
+            {exampleItems.map((element, index) => {
               const {
                 label,
                 action: elementAction,
+                projectNumber,
                 description,
                 deprecated,
               } = element;
               return (
                 <ListItem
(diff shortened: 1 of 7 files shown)


src/shared/components/Animations/ListItem.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+  boolean,
+} from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
+import ListItem from './ListItem';
+import { Tag, Button } from 'antd';
+
+const exampleItems = [
+  {
+    label: 'nexus',
+    projectNumber: 1200,
```
julienmachon: Do we stick with that `projectNumber` thing? I though we scrapped that?

src/shared/components/Animations/ListItem.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import './list-item.less';
+
+export interface ListItemProps {
+  label: string;
+  projectNumber?: number;
```
julienmachon: I'd remove that. Plus I guess this list can be anything, not necessarily an organisation list

src/shared/components/Animations/ListItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+  description?: string;
+  logo?: string;
+  deprecated?: boolean;
+  onClick?(): void;
+  onEdit?(): void;
```
julienmachon: Instead of having `onEdit`, can we think of a more generic way of displaying a menu on an item. That menu could be a simple button (passed as a prop) or a more elaborate one?

Also maybe we should add an `onHover` prop as well. My thinking is that we may not check the permissions of every single items at once, but only `onHover`?

src/shared/components/Animations/ListItem.stories.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <ul>
+            {exampleItems.map(element => {
+              const {
+                label,
+                action: elementAction,
+                description,
+                deprecated,
+              } = element;
+              return (
+                <ListItem
```
julienmachon: Do you not get a warning for not having a `key` in here?

src/shared/components/Animations/ListItem.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import './list-item.less';
+
+export interface ListItemProps {
+  label: string;
+  projectNumber?: number;
+  description?: string;
+  logo?: string;
+  deprecated?: boolean;
```
julienmachon: instead of having a `deprecated` prop, I'd create a `tags` props. An array of [tags](https://ant.design/components/tag/) to display

src/shared/components/Animations/ListItem.tsx
kenjinp: I like the onHover Idea, we can use it in many places, I'll think about adding it separately as a custom hook or something and then adding it afterwards

src/shared/components/Animations/ListItem.tsx
kenjinp: I forgot to change the props from the old component!

I'll keep everything generic here

src/shared/components/Animations/ListItem.tsx
kenjinp: forgot to make this generic

src/shared/components/Animations/ListItem.stories.tsx
kenjinp: I made it mandatory! :)

src/shared/components/Animations/ListItem.tsx
kenjinp: removed `onEdit`, I think its better for the parent component to simply pass in whatever it needs there

src/shared/components/Animations/ListItem.tsx
kenjinp: 👍

src/shared/components/Animations/ListItem.stories.tsx
kenjinp: just an arbitrary example

src/shared/components/Animations/ListItem.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import './list-item.less';
+
+export interface ListItemProps {
+  label: string;
+  key: string;
+  description?: string;
+  detail?: React.ReactComponentElement<any>;
+  action?: React.ReactComponentElement<any>;
+  onClick?(): void;
```
julienmachon: `onClick` should probably return the `id` of the element that was selected?

src/shared/components/Animations/ListItem.stories.tsx
julienmachon: `Warning: ListItem: `key` is not a prop.` This is what I get when I run the storybook. `key` is a reserved prop I think

src/shared/components/Animations/ListItem.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  description={description}
+                  action={
+                    elementAction ? (
+                      <Button
+                        icon="edit"
+                        className="edit-button"
+                        size="small"
+                        type="primary"
+                        tabIndex={1}
+                        onClick={() => action('edit-clicked')}
```
julienmachon: When I click `edit`, I get the `element-clicked` event instead of the `edit-clicked` one

src/shared/components/Animations/ListItem.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import './list-item.less';
+
+export interface ListItemProps {
+  label: string;
+  key: string;
```
julienmachon: `key` is a reserved prop I think, so it cannot be here

src/shared/components/Animations/ListItem.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+                description,
+                deprecated,
+              } = element;
+              return (
+                <ListItem
+                  key={`element-${index}`}
+                  label={label}
+                  onClick={action('element-clicked')}
+                  detail={
+                    <div style={{ display: 'flex' }}>
```
julienmachon: do you think it's ok having to style the container of details? detail prop components are appended to a `.detail` container, so maybe display flex should be set on that container?

I'm also having an internal debate for whether it should be `detail` or `details` :grimacing:

src/shared/components/Animations/ListItem.tsx
kenjinp: that's a great idea

src/shared/components/Animations/ListItem.stories.tsx
kenjinp: oh man this is a tough one

kenjinp: forgot to call `e.stopPropagation()` !

Also don't forget to call the storybook action with `action('edit-clicked')()` or it wont fire in this callback!

src/shared/components/Animations/ListItem.tsx
kenjinp: I used id instead

src/shared/components/Animations/ListItem.stories.tsx
kenjinp: yes `details` is much better. I've also styled the `details` div but here you need it as well because there are two children

PR #186, opened 2019-03-28
262 implicit flow

julienmachon:
This PR implements the implicit flow for nexus web,

The login form creates a new OpenIDConnect `UserManager` based on which realm is selected.
A cookie is created with the `preferred realm` as well as unique key created as follow `nexus__user:issuerUrl:clientId`

This unique key is the name of the cookie which will contain the user session data

When a `UserManager` object is created, a few events are setup in order to load any existing user session, renew expired token and silent renew any expiring tokens, etc... A redux action is fired when an event is triggered.

The final flow looks like this:
Screenshot from 2019-03-27 17-20-52


src/shared/utils/__tests__/index.spec.ts
```diff
@@ -120,4 +121,16 @@ describe('utils functions', () => {
       expect(getLogoutUrl([identities[1]], realms)).toEqual('');
     });
   });
+  describe('hasExpired()', () => {
+    // this is 1987, so unless you time travel and run this test, it will always be in the past
+    const past: number = 536457600;
+    // this is year 4000, si it unless Nexus is still around in 2000 years, this will always be the future
```
kenjinp: lol

julienmachon: :-)

package.json
```diff
@@ -17,13 +17,15 @@
...
-    "antd": "^3.15.0",
     "@bbp/nexus-sdk": "1.0.0-beta.19",
+    "antd": "^3.15.0",
     "codemirror": "^5.44.0",
     "connected-react-router": "^6.3.2",
     "cookie-parser": "^1.4.4",
+    "cookie-storage": "^3.2.0",
     "d3": "^5.9.1",
     "d3-force": "^2.0.0",
+    "debug": "^4.1.1",
```
kenjinp: whats this?

julienmachon: it's gone.

PR #189, opened 2019-03-28
✨FEAT: Infinite scroll

kenjinp:
# Infinite Scrolling!

comes in two flavors!
- click to load more
onClick

- scroll to load more
onscroll

If your list has a scroll bar (eg: has a `max-height` or some such styling), it will automatically load when the user scrolls to the bottom, no configuration required!


src/shared/components/Animations/InfiniteScroll.tsx
```diff
@@ -0,0 +1,125 @@
...
+  itemClassName?: string;
+  makeKey: (item: any, index: number) => string;
+  next: VoidFunction;
+  style?: React.CSSProperties;
+  fetchablePaginatedList: FetchableState<PaginatedList<any>>;
+}
+
+const InfiniteScroll: React.FunctionComponent<InfiniteScrollProps> = props => {
+  const {
+    makeKey,
```
julienmachon: what does this do?

src/shared/components/Animations/InfiniteScroll.tsx
```diff
@@ -0,0 +1,125 @@
...
+    itemClassName,
+    itemComponent,
+    fetchablePaginatedList,
+    next,
```
julienmachon: I have to look in the code to understand what that does. Maybe this could be called `loadMore`?

src/shared/components/Animations/InfiniteScroll.tsx
```diff
@@ -0,0 +1,125 @@
...
+  next: VoidFunction;
+  style?: React.CSSProperties;
+  fetchablePaginatedList: FetchableState<PaginatedList<any>>;
+}
+
+const InfiniteScroll: React.FunctionComponent<InfiniteScrollProps> = props => {
+  const {
+    makeKey,
+    itemClassName,
+    itemComponent,
```
julienmachon: I don't get this prop, what does it do exactly?

src/shared/components/Animations/InfiniteScroll.tsx
kenjinp: yeah, maybe it should be called loadNextPage

src/shared/components/Animations/InfiniteScroll.tsx
kenjinp: it's the list element that will be rendered for each item

src/shared/components/Animations/InfiniteScroll.stories.tsx
```diff
@@ -0,0 +1,359 @@
...
+
+storiesOf('Components/InfiniteScroll/Basic', module)
+  .addDecorator(withKnobs)
+  .add(
+    'InfiniteScroll',
+    withInfo(`
+    Scroll all the way down and ask for more!
+
+    ~~~js
+    <InfiniteScroll
```
julienmachon: I was expecting only 3 props:
- `items` a list of items to render
- `onItemClick(itemID)` callback when an item is clicked
- `onLoadMore()?` optional callback when load more button clicked (if none, no loadmore button)

src/shared/components/Animations/InfiniteScroll.tsx
kenjinp: this is a map function so you can take the item and return a value to make a string for the key, useful I think

src/shared/components/Animations/InfiniteScroll.stories.tsx
```diff
@@ -0,0 +1,362 @@
...
+              next={next}
+              fetchablePaginatedList={fetchablePaginatedList}
+            />
+          </div>
+        );
+      });
+    })
+  );
+
+storiesOf('Components/InfiniteScroll/Simulated/OnClick', module)
```
julienmachon: This is quite a story!! :open_mouth:

src/shared/components/Animations/InfiniteScroll.stories.tsx
kenjinp: The items are agnostic basically so you can pass whatever event handlers you want from the parent

I also though to use the paginated settings conventions we used elsewhere in the app for this. 

This element takes care of handling the state of results, so that each page is concatenated to each other, so the parent doesn't need to do that

src/shared/components/Animations/InfiniteScroll.stories.tsx
kenjinp: Yeah, but I want to make the story page clear except for one example each... not sure another good way to do that

src/shared/components/Animations/InfiniteScroll.tsx
kenjinp: Another reason why we need this is because the elements are animated, and those animated react elements must host the key, not the item component thats rendered inside it, so we need a way to map the keys inside the InfiniteScroll component itself

src/shared/components/Animations/InfiniteScroll.stories.tsx
julienmachon: What I don't get is that there is no pagination in this component, and yet we have to deal with  a `Paginated` list...

kenjinp: Well there is pagination for sure, it's just that it's always sequential.

PR #193, opened 2019-04-02
increase max old space

julienmachon:
It will hopefully fix this
Screenshot from 2019-04-02 14-42-59

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM node:10-alpine as builder
 
 WORKDIR /tmp/nexus-web
 COPY . /tmp/nexus-web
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM node:10-alpine as builder
 
 WORKDIR /tmp/nexus-web
 COPY . /tmp/nexus-web
-RUN npm ci && npm run build
+RUN npm ci && node --max-old-space-size=4096 `which npm` run build
```
mdupont: There should be a comment next to this command to explain why we use max-old-space

julienmachon: Done!

PR #197, opened 2019-04-05
Bugfix/localstorage issue

julienmachon:
- Added the default key to the local storage (needs to come from config, I'll fix that in a separate PR)- 
- Remove local data when logging out
- don't generate dynamic keys when mapping `Lists`
- instead attach unique ID to list on creation
- only pass `List[]` that matches `org/project` rather then the whole list state

### DIFF
--- a/src/shared/views/Project.tsx
+++ b/src/shared/views/Project.tsx
@@ -7,7 +7,7 @@ import { fetchOrg } from '../store/actions/nexus/activeOrg';
 import { Empty, Switch, Icon, Tooltip, Button } from 'antd';
 import Menu from '../components/Workspace/Menu';
 import { createList, initializeProjectList } from '../store/actions/lists';
-import { ListsByProjectState } from '../store/reducers/lists';
+import { List } from '../store/reducers/lists';
 import { Project, Resource, NexusFile } from '@bbp/nexus-sdk';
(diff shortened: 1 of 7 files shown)


src/shared/views/Project.tsx
```diff
@@ -168,26 +168,39 @@ const ProjectView: React.FunctionComponent<ProjectViewProps> = ({
...
-    null,
-  lists: state.lists || {},
-});
+    null;
+
+  const orgLabel = (orgData && orgData.org.label) || '';
+  const projectLabel = (projectData && projectData.label) || '';
+  const activeListId =
+    (state.lists &&
+      Object.keys(state.lists).find(key => key === orgLabel + projectLabel)) ||
```
mdupont: They should be joined with a kind of separator so that there cannot be collisions if an orgLabel is the same than another orgLabel+projectLabel

julienmachon: I don't really want to change too much logic, it's a little brittle plus a list only belongs to a project, that belongs to an org. And you can't have 2 orgs with the same name, so `org`+`project` will always be unique to the specific project

julienmachon:
> Can the state be different depending on the authentication? Anonymous, user 1, user 2... One should never inherit the state of another.

You cannot be 2 users at the same time. When you logout, the local state is deleted.

PR #203, opened 2019-04-15
✨ Feat: Workspace Selector Menu

kenjinp:
Select your orgs and projects from this menu!

### DIFF
--- a/src/shared/components/Header/WorkspaceSelectorMenu/WorkspaceSelectorContainer.tsx
+++ b/src/shared/components/Header/WorkspaceSelectorMenu/WorkspaceSelectorContainer.tsx
@@ -9,6 +9,7 @@ import {
   PaginatedList,
 } from '@bbp/nexus-sdk';
 import { fetchOrgs } from '../../../store/actions/nexus/orgs';
+import { assignOrgAndFetchOrgProjects } from '../../../store/actions/nexus/activeOrg';
 import { FetchableState } from '../../../store/reducers/utils';
 
 export interface WorkspaceSelectorContainerProps {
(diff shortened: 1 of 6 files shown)


src/shared/components/Header/WorkspaceSelectorMenu/WorkspaceSelectorContainer.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const mapStateToProps = (state: RootState) => ({
+  displayPerPage: state.uiSettings.pageSizes.workspaceSelectorMenuSize,
+  orgs: (state.nexus && state.nexus.orgs) || {
+    isFetching: false,
+    data: null,
+    error: null,
+  },
+  activeOrg:
+    (state.nexus &&
```
mdupont: These nested checks are so ugly... I'm far from being a proponent of adding yet another dependency, but we may benefit from using [lodash.get](https://www.npmjs.com/package/lodash.get), which is self-contained, to make this easier to read and less brittle.

src/shared/components/Header/WorkspaceSelectorMenu/WorkspaceSelectorMenuButton.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { Button, Popover } from 'antd';
+import { WorkspaceSelectorContainerProps } from './workspaceSelectorContainer';
+import { Organization, Project } from '@bbp/nexus-sdk';
+import WorkspaceSelectorMenu from './WorkspaceSelectMenu';
+
+const getButtonLabel = (
+  activeOrg: Organization | null,
```
mdupont: Any reason why it's not `activeOrg?: Organization`

kenjinp: I just like null better. I know I put it there

src/shared/components/Header/WorkspaceSelectorMenu/WorkspaceSelectorContainer.tsx
kenjinp: we have a getProp function in the utils folder that can be used for this

PR #208, opened 2019-04-26
👷‍♀️FIX: Resource Searchability

kenjinp:
Refactored Lists to resolve a number bugs and streamline the styling and UX. 

Screenshot 2019-04-26 at 14 09 35

## FEATS
- Resource Query Lists now have an updated filtering panel so you can see what the list is currently filtered by
- Resource Query Lists now use infinite scrolling
- the Queries Board (this main page) now behaves much like trello, with a fixed height and side scrolling
- eliminated the need for some specialised lists, resources are now displayed using generic lists
- results will not be stored in local storage anymore, but the list settings will

## BUGS DEFEATED
- Resource Query Lists now will update independently, performing an action on one list won't cause all the others to refresh
- Text search will be applied along with the filter dropdowns together, and not randomly depending on what you last clicked, or something. 
- Resource Query Lists filter settings will be entirely reset when you click the little "clear filters" button


src/shared/components/Types/TypesIcon.tsx
```diff
@@ -20,14 +20,14 @@ const TypesIcon: React.SFC<TypesIconProps> = ({ type }) => {
   }).toString();
   const src = `data:image/png;base64,${imageData}`;
   return (
-    <li className="types-icon">
```
kenjinp: I had to change to Div, because it's quite possible that LI will exist within LI and thats not HTML5 compliant

src/shared/components/Workspace/Queries/Query/QueryComponent.tsx
```diff
@@ -0,0 +1,178 @@
...
+  return (
+    <div className="query-component">
+      <h3 className="header">
+        <RenameableItem
+          defaultValue={`${name} ${isFetching ? 'loading' : ''}`}
+          onChange={handleUpdate}
+          size="small"
+        />
+        <div className="count">
+          {fetchablePaginatedList.data &&
```
mdupont: Could be much more concise if there were a line before the `return` with `const total = fetchablePaginatedList.data && fetchablePaginatedList.data.total || 0;`

src/shared/components/Workspace/Queries/Query/QueryComponent.tsx
```diff
@@ -0,0 +1,178 @@
...
+                  >
+                    {resource.name}
+                  </Popover>
+                }
+                id={resource.id}
+                details={
+                  resource.type && resource.type.length ? (
+                    <TypesIconList type={resource.type} />
+                  ) : (
+                    undefined
```
mdupont: Bad practice to manually use `undefined`, use `null` instead or an empty JSX element `<></>`

src/shared/components/Workspace/Queries/queries-board.less
```diff
@@ -0,0 +1,22 @@
+.queries-board {
+  // 117 = 40 (header) + 42 (project banner) + 35 (footer)
+  height: calc(100vh - 117px);
```
mdupont: cannot we automate that with Flexbox or CSS Grid on the whole container of header + body + footer?

src/shared/components/Workspace/Queries/Query/filter-dropdown.less
```diff
@@ -0,0 +1,15 @@
...
+  & > .ant-select {
+    width: 100%;
+  }
+}
+
+.drop-option {
+  display: flex;
+  justify-content: space-between;
+  & > .label {
+    margin-right: 1em;
```
mdupont: where does `1em` comes from?

src/shared/components/Workspace/Queries/Query/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const {
+    list: { name, id, results, query },
+    queryResources,
+  } = props;
+  React.useEffect(() => {
+    handleRefreshList();
+  }, [query && query.textQuery, query && query.filters]);
+
+  const handleRefreshList = () => {
+    const size = 20;
```
mdupont: Magic number, we have a constant for that

src/shared/components/Workspace/Queries/Query/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    queryResources(id, { size, from: 0 }, query);
+  };
+
+  const next = () => {
+    const size = 20;
```
mdupont: Same, magic number

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
+@import '../../../../lib.less';
+
+.query-component {
+  .cardlike();
+  background-color: #f5f6f7;
```
mdupont: Should be a less variable

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
...
+  width: 400px;
```
mdupont: Should be a variable in lib.less

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
...
+    margin-left: 1em;
+  }
+  & > .controls {
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: 1em;
+    flex-wrap: wrap;
+  }
+  & > .controls > .search {
+    width: 60%;
```
mdupont: Mixing ems, rems, pxs and percents... Also saw viewport units (vh / vw) elsewhere. Confusing to me

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
...
+    width: calc(50% - 2px);
+    margin: 0.5em 0;
+  }
+  & > .infinite-scroll > .list .list-item {
+    margin: 0.5em 0;
+  }
+  &.-add {
+    cursor: pointer;
+    padding: 1em;
+    color: hsla(0, 9%, 91%, 0.8);
```
mdupont: should be a variable. Also, why hsla() here instead of rgba() which is closer to the hex notation used elsewhere?

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
...
+    transition: all 85ms ease-in;
```
mdupont: timing should be a variable

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
...
+    background-color: #00000050;
```
mdupont: Should be a variable

src/shared/components/Workspace/Queries/Query/query-component.less
```diff
@@ -0,0 +1,55 @@
...
+    height: fit-content;
+    display: block;
+    &:hover {
+      background-color: #00000070;
```
mdupont: Guess what? :-D

src/shared/styles/project-banner.less
```diff
@@ -0,0 +1,22 @@
+@import '../lib.less';
+
+.project-banner {
+  background-color: #00000005;
```
mdupont: bla bla variable. Would reduce slight variations

src/shared/components/Workspace/Queries/queries-board.less
kenjinp: There's probably a way to do that but it will require some learning on my part

src/shared/components/Workspace/Queries/Query/query-component.less
kenjinp: these percentages are for taking up width on a relative flexbox, just to look nice on the control panel.

 vh and vw are only for components that are layouts, basically, I think I have used those sparingly.

pxs should not really ever be used, instead em I guess

src/shared/components/Workspace/Queries/queries-board.less
kenjinp: I'll experiment with flexbox and flex-grow to get this effect

kenjinp: I'll leave this to another patch because those changes will reorder the other pages that we don't have consistent solutions for yet, especially the org lists and such.

PR #215, opened 2019-05-03
👷‍♀️FIX: list perceived speed improvements

kenjinp:
# NEW
Bug fixes and changes to further improve UX while working with resources and especially improving the perceived speed of query lists

## Changes / Features
- changed `@id` and `_self` copy to `Identifier` and `Nexus Address` respectively based on @MFSY 's sage advice
- new gradient color on query list titles while loading to make it more obvious why you're waiting. 

## Bugfixes
- fixes a problem where switching tabs would cause the resource editor to not be editable. 
- removes react-spring animations from lists because y-frame redraws were too costly. 
- adds debouncing to `useMeasure()` reading to reduce content height measurements
- [implements `use passive`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) on useInfiniteScroll for performance on `addEventListener`


src/shared/utils/debounce.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
mdupont: Why not use the npm package?

src/shared/components/hooks/useInfiniteScroll.ts
```diff
@@ -10,22 +12,32 @@ const useInfiniteScroll = (
   total: number
 ) => {
   const ref = React.useRef<HTMLDivElement>(null);
+
   if (!isBrowser) {
     return [{ ref }];
   }
 
   React.useEffect(() => {
+    const debounceMS = 100;
```
mdupont: Magic number

src/shared/utils/debounce.ts
kenjinp: I think I'll have to because it's making typescript fail storybook compilation :O

PR #243, opened 2019-06-13
👷‍♀️FIX: better full text search support

kenjinp:
resolves https://github.com/BlueBrain/nexus/issues/663

This is a partial victory, there are some cases where the matches still cannot be found!

To be further investigated with @umbreak

### DIFF
--- a/src/shared/store/actions/utils/makeESQuery.ts
+++ b/src/shared/store/actions/utils/makeESQuery.ts
@@ -56,7 +56,7 @@ export const makeESQuery = (query?: List['query']) => {
     }
     if (query.textQuery) {
       must.push({
-        // from https://github.com/BlueBrain/nexus-kg/pull/675/files#diff-4aa2f8ee4ce4694213bc1e9add6895c0R9
+        // from https://github.com/BlueBrain/nexus-kg/blob/master/src/test/resources/search/query-schema-rev-createdBy-q.json#L10-L12
         match: {
           _all_fields: query.textQuery,
         },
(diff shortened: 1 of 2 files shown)


src/shared/store/actions/utils/makeESQuery.ts
```diff
@@ -56,8 +56,8 @@ export const makeESQuery = (query?: List['query']) => {
     }
     if (query.textQuery) {
       must.push({
-        query_string: {
-          query: `${query.textQuery}~`,
+        match: {
```
mdupont: Could you add a comment to explain why you are using `match` here?

kenjinp: I added it. It's from the implementation of the full_text_search query param on KG. https://github.com/BlueBrain/nexus-kg/pull/675/files#diff-4aa2f8ee4ce4694213bc1e9add6895c0R9

kenjinp:
Actually after some testing it seems the query works very well and I cannot find examples where it doesn't work as expected :)

kenjinp:
The full text search query seems to be a bit strict though...

julienmachon:
I don't think that make any difference at all!! haha, but I guess it is better to use whatever they do in KG

src/shared/store/actions/utils/makeESQuery.ts
julienmachon: Not sure the comment is 100% needed in the code itself but if you decide to keep it there, can you link to the actual file rather that the PR: https://github.com/BlueBrain/nexus-kg/blob/master/src/test/resources/search/query-schema-rev-createdBy-q.json#L10-L12

kenjinp:
¯\_(ツ)_/¯

PR #244, opened 2019-06-14
👷‍♀️FIX: 661 file upload links

kenjinp:
- Fixes a bug where posting files that have failed would not appear as a failure
- Provides a nice link after posting a file for the user to navigate to their newly created resource

resolves https://github.com/BlueBrain/nexus/issues/661


src/client/index.tsx
```diff
@@ -122,9 +123,11 @@ const setupUserSession = async (userManager: UserManager, store: Store) => {
 const renderApp = () => {
   return ReactDOM.render(
     <Provider store={store}>
-      <ConnectedRouter history={history}>
-        <App />
-      </ConnectedRouter>
+      <NexusContext.Provider value={nexus}>
```
julienmachon: I feel really uncomfortable doing this, especially when there is already a redux action setup for files. I'm ok to sit down together on this one whenever you're free

PR #246, opened 2019-06-14
👷‍♀️FIX: file upload error

kenjinp:
Fixes a bug where errors would not show up after an upload failed
Screenshot 2019-06-14 at 11 58 58

### DIFF
--- a/src/shared/views/Project.tsx
+++ b/src/shared/views/Project.tsx
@@ -260,9 +260,8 @@ const mapDispatchToProps = (dispatch: any) => {
         payload
       );
     },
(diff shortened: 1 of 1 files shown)


src/shared/views/Project.tsx
```diff
@@ -260,9 +260,8 @@ const mapDispatchToProps = (dispatch: any) => {
         payload
       );
     },
-    createFile: async (file: File, options?: CreateFileOptions) => {
-      dispatch(createFile(file, options));
-    },
+    createFile: (file: File, options?: CreateFileOptions) =>
```
julienmachon: :laughing:

PR #253, opened 2019-06-17
✨ FEAT: User page

kenjinp:
Adds a user page at `/user` that displays your `Identities`
Also cleaned up the `*-view` css classes


kenjinp:
Note, if you're not logged in it will not show you a link to this user page, but you can still navigate there manually for debugging purposes

Screenshot 2019-06-17 at 15 22 03

PR #260, opened 2019-06-19
✨ FEAT: Frequented Projects

kenjinp:
Makes navigating around the app snappier by
- adding frequented projects to a list on the org/project selection screen
- using connected-router to navigate when clicking on the N on the top to go to the home page

This is real big stuff, people. 

Screenshot 2019-06-19 at 11 56 28

### DIFF
--- a/src/shared/components/Animations/ListItem.tsx
+++ b/src/shared/components/Animations/ListItem.tsx
@@ -12,6 +12,7 @@ export interface ListItemProps {
   preview?: ImagePreviewProps;
   action?: React.ReactComponentElement<any> | null;
   onClick?: (id: string, event: React.MouseEvent) => void;
+  onAllClick?: (id: string, event: React.MouseEvent) => void;
   popover?: PopoverProps;
 }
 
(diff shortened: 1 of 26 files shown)


src/shared/components/Animations/ListItem.tsx
```diff
@@ -24,13 +25,23 @@ const ListItem: React.FunctionComponent<ListItemProps> = ({
   action,
   preview,
   popover,
+  onAllClick,
```
mdupont: naming: I have no Idea what 'onAllClick' means. Is there a link labeled 'All'? or is it an event that will be launched on all links?

src/shared/components/Header/__tests__/__snapshots__/Header.test.tsx.snap
```diff
@@ -60,7 +60,11 @@ exports[`Header component Should render correctly 1`] = `
       mouseEnterDelay={0.15}
       mouseLeaveDelay={0.1}
       overlay={
-        <Menu>
+        <Menu
+          className=""
```
mdupont: since this `className` is empty, can it be removed?

src/shared/components/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+
+const useLocalStorage = (key: string) => {
```
mdupont: Love it 😍

src/shared/components/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,14 @@
...
+  const [value, setValue] = React.useState<any>({});
+  React.useEffect(() => {
+    setValue(JSON.parse(localStorage.getItem(key) || '{}'));
```
mdupont: since localStorage may not be available if the code is run on Node.js (unless they install a package to emulate it), maybe there should be a check about whether it is present or not

src/shared/components/Header/__tests__/__snapshots__/Header.test.tsx.snap
kenjinp: This was autogenerated from the snap

src/shared/components/hooks/useLocalStorage.ts
kenjinp: 😘

src/shared/components/Animations/ListItem.tsx
kenjinp: I decided to swap it with the other. so now there's `onClick` and `onTitleClick`

PR #261, opened 2019-06-19
Improve resource creation

julienmachon:
I'm sure there is much more we can do, but this is a start (plus I don't want to add too much functionality until we migrate to the new SDK)

Screenshot from 2019-06-19 15-00-38
Screenshot from 2019-06-19 15-00-47


julienmachon:
I you have an idea for hidding this annoying "expanded" button on resource creation, I take it :wink:

kenjinp:
`showExpanded: false`

PR #264, opened 2019-06-19
✨FEAT: History

kenjinp:
A new tab to checkout changes to a resource over time

Screenshot 2019-06-19 at 17 25 32


src/shared/store/actions/nexus/resource.ts
```diff
@@ -85,6 +85,8 @@ export const fetchAndAssignResource: ActionCreator<ThunkAction> = (
         resourceId,
         {
           expanded,
+          // @ts-ignore
+          rev: 2,
```
julienmachon: :gun:

PR #274, opened 2019-06-25
👷‍♀️ FIX: deprecate dangerously action

kenjinp:


kenjinp:
This actions are built using the props and the keys of that object. it won't know to use the proper method unless you declare it as an argument to `makeActions`

src/shared/components/Resources/ResourceActions.tsx
```diff
@@ -121,6 +121,7 @@ export interface ResourceActionsProps {
   goToElasticSearchView: (resource: Resource) => void;
   deprecateResource: (resource: Resource) => void;
   downloadFile: (resource: Resource) => void;
+  veryDangerousToDeprecate: (resource: Resource) => void;
```
julienmachon: What's the difference between `deprecateResource` and `veryDangerousToDeprecate` ?

kenjinp: one is the normal message, "are you sure?", and the other is "this will break stuff!"

julienmachon: I mean... we're not going to add a new prop everytime we need a different message right? can we not do something like:
```suggestion
  deprecateResource: (resource: Resource, message?: string) => void;
```

kenjinp: You're right, I fixed the actions to be a little more generic, now you can create a button that uses another action and you dont have to add a new prop type every time

julienmachon:
Let's merge this for now with the idea in mind that we will refactor soon: https://github.com/BlueBrain/nexus/issues/688

PR #275, opened 2019-06-26
✨ FEAT: 681 sort query by creation date

kenjinp:
resolves : https://github.com/BlueBrain/nexus/issues/681


kenjinp:
That was a great point, but it turns out I goofed up because each entry is supposed to be its own object. 

Also note for the future, `Set` serializes into an empty object in JSON so it wouldn't work for this case anyways.

PR #276, opened 2019-06-26
👷‍♀️ FIX: add types to the pre-loaded payloads

kenjinp:
resolves: https://github.com/BlueBrain/nexus/issues/678

### DIFF
--- a/src/shared/components/Resources/defaultResourcePayloads.ts
+++ b/src/shared/components/Resources/defaultResourcePayloads.ts
@@ -7,7 +7,7 @@ export const DEFAULT_RESOURCE = {
 export const DISK_STORAGE = {
   '@context': {}, // optional
   '@id': '', // optional
-  '@type': ['DiskStorage'],
+  '@type': 'DiskStorage',
   default: false,
   volume: '',
   readPermission: [], // optional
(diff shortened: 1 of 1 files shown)


src/shared/components/Resources/defaultResourcePayloads.ts
```diff
@@ -7,6 +7,7 @@ export const DEFAULT_RESOURCE = {
 export const DISK_STORAGE = {
   '@context': {}, // optional
   '@id': '', // optional
+  '@type': ['DiskStorage'],
```
julienmachon: Documentation says `@type: "DiskStorage"`. I know this works too but we might as well be consistent with the doc

src/shared/components/Resources/defaultResourcePayloads.ts
```diff
@@ -36,6 +39,7 @@ export const S3_STORAGE = {
 export const SPARQL_VIEW = {
   '@context': {}, // optional
   '@id': '', // optional
+  '@type': ['View', 'SparqlView'],
```
julienmachon: I would just do 
```suggestion
  '@type': 'SparqlView',
```

src/shared/components/Resources/defaultResourcePayloads.ts
```diff
@@ -57,18 +62,21 @@ export const ELASTIC_SEARCH_VIEW = {
...
   '@context': {}, // optional
   '@id': '', // optional
+  '@type': ['View', 'AggregateElasticSearchView'],
   views: [{ project: '', viewId: '' }],
 };
 
 export const IN_PROJECT = {
   '@context': {}, // optional
   '@id': '', // optional
+  '@type': ['InProject', 'Resolver'],
```
julienmachon: I'd remove `Resolver` here as well.

PR #277, opened 2019-06-27
👷‍♀️ FIX: update routed menu to use react-router

kenjinp:
Uses react-router for cool points. 
Is no longer animated, as react-spring made it quite laggy


julienmachon:
Is it possible to have either a comment at the top of your component showing how it works or even better, a story in storybook?

kenjinp:
It has a storybook already :)

PR #279, opened 2019-06-27
👷‍♀️FIX: Org Item improvements

kenjinp:
Trying to remove some code and styling overhead by using the `ListItem` component when display orgs. It's now just a thin wrapper, instead of a bespoke styled component. The use should be exactly the same as previously. 

Renamed `OrgCard` to `OrgItem`

Removed the deprecated tags, as we're to filter deprecated orgs and projects out.


src/shared/components/Orgs/OrgItem.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react';
+import { Button } from 'antd';
+import { AccessControl } from '@bbp/react-nexus';
+import ListItem from '../Animations/ListItem';
+
+export interface OrgItemProps {
+  label: string;
+  description?: string;
+  numProjects?: number;
```
julienmachon: I don't like this props name... Is this the number of projects? I would call it `numberOfProjects`

src/shared/components/Orgs/OrgList.tsx
```diff
@@ -55,7 +55,7 @@ const OrgList: React.FunctionComponent<OrgListProps> = ({
       /> */}
       <AnimatedList
```
julienmachon: Can we remove this as well?

kenjinp: I'll go over the OrgList in the next PR

PR #280, opened 2019-06-27
👷‍♀️ FIX: Improve orgs list

kenjinp:
New Org List that can be used standalone. Uses its own fetching state, and exports both a stateless component and a stateful fetching container, the default being the whole container.

It takes no mandatory properties so you can drop this anywhere and get going

infinitescrollorgs

### DIFF
--- a/src/shared/components/Orgs/OrgList.tsx
+++ b/src/shared/components/Orgs/OrgList.tsx
@@ -1,83 +1,162 @@
 import * as React from 'react';
-import { Input, Empty } from 'antd';
-import OrgItem, { OrgItemProps } from './OrgItem';
-import AnimatedList from '../Animations/AnimatedList';
-
+import { Empty, Pagination, Spin, Button } from 'antd';
+import { OrgResponseCommon, OrganizationList } from '@bbp/nexus-sdk';
+import { PaginationSettings } from '../../utils/types';
+import Search from 'antd/lib/input/Search';
+import OrgItem from './OrgItem';
+import { AccessControl, useNexus } from '@bbp/react-nexus';
 import './Orgs.less';
-import { Organization } from '@bbp/nexus-sdk-legacy';
 
-export interface OrgListProps {
-  orgs: OrgItemProps[];
-  busy?: boolean;
-  error?: { message: string; name: string };
-  onOrgClick?(org: Organization): void;
-  onOrgEdit?(label: string): void;
-  paginationSettings?: { total: number; from: number; pageSize: number };
-  onPaginationChange?: (page: number, pageSize?: number) => void;
+export interface OrgListContainerProps {
+  pageSize: number;
+  onOrgClick?(orgLabel: string): void;
+  onOrgEdit?(orgLabel: string): void;
+  createOrg?(): void;
 }
 
-const Search = Input.Search;
+export const OrgsListContainer: React.FunctionComponent<
+  OrgListContainerProps
+> = props => {
+  const { pageSize = 20, onOrgClick, onOrgEdit, createOrg } = props;
+  const [searchValue, setSearchValue] = React.useState<string>();
+  const [{ from, size }, setPagination] = React.useState({
+    size: pageSize,
+    from: 0,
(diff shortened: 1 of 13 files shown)


src/shared/components/Orgs/OrgList.tsx
```diff
@@ -1,83 +1,162 @@
...
+};
+
+export interface OrgListProps {
+  orgs: OrgResponseCommon[];
+  loading?: boolean;
+  error?: Error;
+  onOrgClick?(orgLabel: string): void;
+  onOrgEdit?(orgLabel: string): void;
+  createOrg?(): void;
+  paginationSettings: PaginationSettings;
```
julienmachon: I would get rid of pagination and so `load more` like trello for example

src/shared/components/Orgs/OrgList.tsx
```diff
@@ -1,83 +1,162 @@
...
+  onPaginationChange?(page: number, pageSize?: number): void;
```
julienmachon: this would become `onLoadMore()`

src/shared/components/Orgs/OrgList.tsx
```diff
@@ -1,83 +1,162 @@
...
+  onOrgClick?(orgLabel: string): void;
+  onOrgEdit?(orgLabel: string): void;
+  createOrg?(): void;
 }
 
-const Search = Input.Search;
+export const OrgsListContainer: React.FunctionComponent<
+  OrgListContainerProps
+> = props => {
+  const { pageSize = 20, onOrgClick, onOrgEdit, createOrg } = props;
```
julienmachon: We need to use the same default value everywhere (from a exported const)

kenjinp: I used the default UI config state const, but I also broke it into a new file in `utils/consts` where I hope we can find these things easier

src/shared/components/Orgs/OrgList.tsx
kenjinp: > like trello
🙉

src/shared/components/Orgs/OrgList.tsx
```diff
@@ -1,83 +1,152 @@
...
+    createOrg,
+  } = props;
+  const [{ from, size, label, deprecated }, setQuery] = React.useState({
+    size: pageSize,
+    from: 0,
+    label: '',
+    deprecated: false,
+  });
+  const [fetchablePaginatedList, setPaginatedList] = React.useState<
+    FetchableState<PaginatedList<OrgResponseCommon>>
```
julienmachon: I'm not sure I get that type. Why do we need a `FetchableState of PaginatedList of OrgResponseCommon`. What about `OrganisationList` directly?

src/shared/components/Orgs/OrgList.tsx
```diff
@@ -1,83 +1,152 @@
...
+      nexus.Organization.list({
+        from,
+        size,
+        label,
+        deprecated,
+      }),
+    [from, size, label, deprecated]
+  );
+
+  React.useEffect(() => {
```
julienmachon: I'm not sure I get that bit. You already have all the data you need above. All you're doing in that `useEffect` is derive a new state by combining 2, without even changing any values. :thinking:

src/shared/components/Orgs/OrgList.tsx
kenjinp: The InfiniteScroll component requires it in this format, it was the standard until we changed the SDK so it made sense at the time.

src/shared/components/Orgs/OrgList.tsx
kenjinp: Yeah this is one I struggled with. The problem is we only want to update the InfiniteScroll list when the response data comes, and not when the state changes, otherwise InfiniteScroll won't know if it's a new list and won't refresh itself. 

Maybe extending InfiniteScroll with a refresh function prop to force a refresh would be a way to avoid this.

PR #281, opened 2019-06-28
Feat/view statistics progress

julienmachon:
Display the status of a Nexus View.

As the Developer screen in our App is powered by the default Nexus ElasticSearch View, we now display a progress gauge in the banner.

Screenshot 2019-06-28 at 15 01 30

We also display the progress of any view in the Query page:
Screenshot from 2019-07-04 09-21-46
Screenshot from 2019-07-04 09-22-00
Screenshot from 2019-07-04 09-22-07


src/shared/utils/index.ts
```diff
@@ -204,3 +204,51 @@ export function blacklistKeys(raw: { [key: string]: any }, keys: string[]) {
...
+    givenDate.getDay() === today.getDay() &&
+    givenDate.getMonth() === today.getMonth() &&
+    givenDate.getFullYear() === today.getFullYear()
+  );
+}
+
+/**
+ * How long ago from today in days
+ *
+ * return -1 if given date if from the future
```
mdupont: Typo: "if from" => "is from"

src/shared/utils/index.ts
```diff
@@ -204,3 +204,51 @@ export function blacklistKeys(raw: { [key: string]: any }, keys: string[]) {
...
+ *
+ */
+export function howLongAgo(date: string): number {
+  const today = new Date(Date.now());
+  const givenDate = new Date(date);
+  const millisecondsInDay = 24 * 60 * 60 * 1000;
```
mdupont: Comparing dates can be tricky (with TZ, leap seconds, DST, etc). Also, we already have momentjs (or a similar lib) in the indirect dependencies, why not use it here to compare date cleanly?

src/shared/utils/index.ts
```diff
@@ -204,3 +204,51 @@ export function blacklistKeys(raw: { [key: string]: any }, keys: string[]) {
...
+  );
+}
+
+/**
+ * How long ago from today in days
+ *
+ * return -1 if given date if from the future
+ *
+ */
+export function howLongAgo(date: string): number {
```
mdupont: Since it returns days, it would be more explicit to call it `howManyDaysAgo()`

src/shared/utils/index.ts
```diff
@@ -204,3 +204,51 @@ export function blacklistKeys(raw: { [key: string]: any }, keys: string[]) {
...
+ */
+export function isToday(date: string): boolean {
+  const today = new Date(Date.now());
+  const givenDate = new Date(date);
+
+  if (givenDate.toString() === 'Invalid Date') {
+    return false;
+  }
+  return (
+    givenDate.getDay() === today.getDay() &&
```
mdupont: This may return incorrect results if Date.now() and the input date have different timezones, better use the UTC versions: `getUTCDay(), getUTCMonth(), getUTCFullYear()` OR make sure that both dates are converted to UTC earlier in the function

src/shared/views/Project.tsx
```diff
@@ -161,8 +162,16 @@ const ProjectView: React.FunctionComponent<ProjectViewProps> = ({
...
                   )}{' '}
-                  {project.label}{' '}
+                  {project.label}
+                  {'  '}
                 </h1>
+                <div style={{ marginLeft: 10 }}>
+                  <ViewStatisticsProgress
+                    orgLabel={(org && org.label) || ''}
+                    projectLabel={project.label}
+                    resourceId="nxv:defaultElasticSearchIndex"
```
mdupont: Pls use the constant we have for this id in another file, instead of a magic string

src/shared/utils/index.ts
```diff
@@ -204,3 +204,51 @@ export function blacklistKeys(raw: { [key: string]: any }, keys: string[]) {
...
+  const millisecondsInDay = 24 * 60 * 60 * 1000;
+
+  // reset hours for an accurate comparison
+  today.setHours(0, 0, 0, 0);
+  givenDate.setHours(0, 0, 0, 0);
+  // + operator to make date a number
+  return (+today - +givenDate) / millisecondsInDay;
+}
+
+export function displayTimeFromDate(date: string): string {
```
mdupont: Do you want to display time in UTC, or in the local timezone of the user (when it's server-side rendered, it will be the timezone of the server), or in the timezone of whatever was stored in Nexus for this resource?

src/shared/components/Views/ViewStatisticsProgress.tsx
```diff
@@ -0,0 +1,71 @@
...
+      return `${daysSinceUpdate} day ago`;
+    default:
+      return `${daysSinceUpdate} days ago`;
+  }
+}
+
+type ViewStatisticsProgressProps = {
+  processedEvents: number;
+  totalEvents: number;
+  lastIndexed: string; // UTC Date
```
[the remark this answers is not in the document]
julienmachon: it's the string coming from nexus. If we day it has to be a date then the props will have to be `new Date(lastIndexed)`

julienmachon:
I resolve most problems by simply using moment.js :+1:

src/shared/views/Project.tsx
julienmachon: we don't have an exported constant :man_shrugging: 

I'll create one

julienmachon: here we go 16c8a07b7c7e57f6ccc9311ab1c7afef398f460d

kenjinp: https://github.com/BlueBrain/nexus-sdk-js/pull/186

PR #284, opened 2019-07-04
👷‍♀️ FIX: filter keys from resource payloads

kenjinp:
resolves this issue: https://github.com/BlueBrain/nexus/issues/691


kenjinp:
This PR is blocked and cannot be built because it requires: https://github.com/BlueBrain/nexus-sdk-js/pull/197

umbreak:
@kenjinp note that this is not required anymore. In order to perform updates, first retrieve the original payload using the .../source endpoint. This won't contain the metadata, so you won't have to manually remove it

PR #285, opened 2019-07-05
Feat/infinite scroll projects

julienmachon:
New `ListItem` component that has styles of any items in a list as well as logic for display actions `onHover` and callback `onClick`

`OrgItem` component that renders one org

`ProjectItem` component that renders one project

Finally, Orgs page and Project page are both displaying a list of `<ListItem><OrgItem /></ListItem>` and `<ListItem><ProjectItem /></ListItem>` respectively


julienmachon:
@kenjinp could you give me a hand with the styles? 

Basically, `ListItem` should be able to render `children` and `actions` on the same line, with action being on the right

and both OrgItem and ProjectItem could do with a little magic. It's quite rough just now.

PR #288, opened 2019-07-05
👷‍♀️ FIX: aggregate query buttons and rename "chainPredicate" to "and"

kenjinp:
For Aggregate Views, those little query buttons wouldn't show up. Now they will!

changed some names because they make much more sense now

### DIFF
--- a/src/shared/utils/nexus-maybe.ts
+++ b/src/shared/utils/nexus-maybe.ts
@@ -3,8 +3,9 @@ import { Resource } from '@bbp/nexus-sdk-legacy';
 const NEXUS_FILE_TYPE = 'File';
 const VIEW = 'View';
 const ES_VIEW = 'ElasticSearchView';
+const AGGREGATE_ES_VIEW = 'AggregateElasticSearchView';
+const AGGREGATE_SPARQL_VIEW = 'AggregateSparqlView';
 const SPARQL_VIEW = 'SparqlView';
-
 const DEFAULT_ELASTIC_SEARCH_VIEW_ID = 'nxv:defaultElasticSearchIndex';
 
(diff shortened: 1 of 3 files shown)


kenjinp:
I also fixed the styling for query view results, and querying Aggregated Views now works too

src/shared/utils/nexus-maybe.ts
```diff
@@ -3,8 +3,9 @@ import { Resource } from '@bbp/nexus-sdk-legacy';
 const NEXUS_FILE_TYPE = 'File';
 const VIEW = 'View';
 const ES_VIEW = 'ElasticSearchView';
+const AGGREGATE_ES_VIEW = 'AggregateElasticSearchView';
+const AGGREGATE_SPARQL_VIEW = 'AggregateSparqlView';
 const SPARQL_VIEW = 'SparqlView';
-
 const DEFAULT_ELASTIC_SEARCH_VIEW_ID = 'nxv:defaultElasticSearchIndex';
 
 const isOfType = (type: string) => (resource: Resource) =>
```
julienmachon: Why is it still using `Resource` from the legacy sdk?

kenjinp: it was made before the conversion to v1.1. Should it be any object, or a new ResourceResponseCommon or so?

julienmachon: Yes, either you create a new type or you use the new SDK one. But if you add new functionality or update something in general, make sure you refactor and remove the legacy dependencies.

kenjinp: Also the views that is using this have the resources in the format such as resource[`@type`] is actually resource.type, so without changing that up the chain for the Resource view we'll have problems using this.

PR #272, opened 2019-06-24
✨ FEAT: Nav menu

kenjinp:
Adds a side menu to navigate between orgs / projects

Adds an `Org` and `Projects` component that each wrap the new sdk for `List` and `Get` methods

Screenshot 2019-06-25 at 10 04 26
Screenshot 2019-06-25 at 10 04 09

PR #290, opened 2019-07-08
✨ FEAT: Nav drawer

kenjinp:
Simplified implementation of https://github.com/BlueBrain/nexus-web/pull/272

Screenshot 2019-07-08 at 14 33 36

### DIFF
--- a/src/shared/App.less
+++ b/src/shared/App.less
@@ -8,6 +8,12 @@
   max-width: @maxConstrainedItemWidth;
   margin: 0 auto;
   display: flex;
+
+  .recently-visited {
+    padding: @default-pad;
+    border: 1px solid @border-color-base;
+    border-radius: @border-radius-base;
+  }
 }
 
 .user-view {
--- a/src/shared/App.less
+++ b/src/shared/App.less
@@ -13,6 +13,7 @@
     padding: @default-pad;
     border: 1px solid @border-color-base;
     border-radius: @border-radius-base;
+    min-height: 400px;
   }
 }
 
--- a/src/shared/components/Menu/NavDrawer.tsx
+++ b/src/shared/components/Menu/NavDrawer.tsx
@@ -1,0 +1,46 @@
+import * as React from 'react';
+import RoutedComponent, { Route } from './RoutedComponent';
+import SideMenu from './SideMenu';
+import './NavDrawer.less';
+
+interface NavDrawerComponentProps {
+  render: (
+    visible: boolean,
+    toggleVisibility: () => void
+  ) => React.ReactElement<any>;
+  routes: Route[];
+  defaultVisibility?: boolean;
+}
+
+export const NavDrawer: React.FunctionComponent<
+  NavDrawerComponentProps
+> = props => {
+  const { render, defaultVisibility, routes } = props;
+
+  const [visible, setVisible] = React.useState(
+    typeof defaultVisibility !== 'undefined' ? defaultVisibility : true
+  );
+
+  const toggleVisibility = () => {
+    setVisible(!visible);
+  };
+  const renderable = render(visible, toggleVisibility);
+
+  return (
+    <>
+      {renderable}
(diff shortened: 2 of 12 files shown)


src/shared/components/Menu/NavDrawer.tsx
```diff
@@ -0,0 +1,46 @@
...
+  defaultVisibility?: boolean;
+}
+
+export const NavDrawer: React.FunctionComponent<
+  NavDrawerComponentProps
+> = props => {
+  const { render, defaultVisibility, routes } = props;
+
+  const [visible, setVisible] = React.useState(
+    typeof defaultVisibility !== 'undefined' ? defaultVisibility : true
```
mdupont: Since `visible` has to be a Boolean, why not write the more explicit `typeof defaultVisibility === boolean`?

src/shared/App.less
```diff
@@ -8,6 +8,13 @@
   max-width: @maxConstrainedItemWidth;
   margin: 0 auto;
   display: flex;
+
+  .recently-visited {
+    padding: @default-pad;
+    border: 1px solid @border-color-base;
+    border-radius: @border-radius-base;
+    min-height: 400px;
```
mdupont: It looks weird when empty with this min-height

src/shared/components/Menu/NavDrawer.tsx
julienmachon: hmm, I would keep it simple and just do:
```suggestion
    defaultVisibility || true
```

src/shared/layouts/MainLayout.tsx
```diff
@@ -86,7 +96,99 @@ const MainLayout: React.FunctionComponent<MainLayoutProps> = ({
...
+                        ))
+                      }
+                    </ProjectList>
+                  </div>
+                );
+              },
+            },
+          ]}
+          defaultVisibility={false}
+          render={(visible, toggleVisibility) => {
```
julienmachon: I had to come down here in order to understand this prop 😝 
I would just remove it. Why not simply have a `visible: boolean` prop instead?
Like these guys: https://ant.design/components/drawer/

src/shared/components/Menu/NavDrawer.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import RoutedComponent, { Route } from './RoutedComponent';
+import SideMenu from './SideMenu';
+import './NavDrawer.less';
+
+interface NavDrawerComponentProps {
+  render: (
```
julienmachon: 🤔

src/shared/App.less
kenjinp: I disagree! the alternative would to have it disappear and that causes jank

src/shared/layouts/MainLayout.tsx
kenjinp: because I wanted to move the toggle logic to the component. You always need a toggle and it's a pain to include toggle state handling everywhere you use such a thing, no?

src/shared/components/Menu/NavDrawer.tsx
kenjinp: its no good?

src/shared/layouts/MainLayout.tsx
julienmachon: Well, I don't find this:

```tsx
const [visible, setVisible] = useState(false);
return (
  <Button
    icon="project"
    size="small"
    onClick={() => setVisible(!visible)}
  >
    Projects
  </Button>
 );
```
to be more painful than:

```tsx
render={(visible, toggleVisibility) => {
            return (
              <Button
                icon="project"
                size="small"
                onClick={() => toggleVisibility()}
              >
                NavDrawer
              </Button>
            );
          }}
```

Also, what if I don't need a toggle component and I just want it to be visible at all time?

PR #291, opened 2019-07-09
POC: Search view

kenjinp:
Example of a `SearchViewConfig` that's required.
```json
{
  "@type": ["SearchViewConfig"],
  "view": "myTargetAggregatedView_SelfURL",
  "facets": "SOME SPARQL QUERY",
  "datasetsQueryConfig": {
    
  }
}
```

You also need an `aggregatedSparqlView`

### DIFF
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -74,6 +74,7 @@ const nexusLegacy = new Nexus({
 });
 Nexus.setEnvironment(preloadedState.config.apiEndpoint);
 // create redux store
+// @ts-ignore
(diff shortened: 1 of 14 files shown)


src/client/index.tsx
```diff
@@ -74,6 +74,7 @@ const nexusLegacy = new Nexus({
 });
 Nexus.setEnvironment(preloadedState.config.apiEndpoint);
 // create redux store
+// @ts-ignore
 const store = configureStore(history, { nexusLegacy, nexus }, initialState);
```
kenjinp: not sure why my build broke without this

PR #322, opened 2019-10-29
Display Nexus Service Versions

dhaneshnm:
Fetch and Display Nexus Service versions on information menu in the
Header.
implements BlueBrain/nexus#773

### DIFF
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -69,11 +69,9 @@ const nexus = createNexusClient({
   uri: preloadedState.config.apiEndpoint,
   links: [setToken],
 });
-
 const nexusLegacy = new Nexus({
   environment: preloadedState.config.apiEndpoint,
 });
-
 Nexus.setEnvironment(preloadedState.config.apiEndpoint);
 // create redux store
 const store = configureStore(history, { nexusLegacy, nexus }, initialState);
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -69,9 +69,11 @@ const nexus = createNexusClient({
   uri: preloadedState.config.apiEndpoint,
   links: [setToken],
 });
+
 const nexusLegacy = new Nexus({
   environment: preloadedState.config.apiEndpoint,
 });
+
 Nexus.setEnvironment(preloadedState.config.apiEndpoint);
 // create redux store
 const store = configureStore(history, { nexusLegacy, nexus }, initialState);
--- a/src/shared/components/Header/__tests__/Header.test.tsx
+++ b/src/shared/components/Header/__tests__/Header.test.tsx
@@ -1,14 +1,14 @@
 import * as React from 'react';
 import { shallow, mount } from 'enzyme';
-import Header, {  serviceVersions } from '..';
+import Header, { ServiceVersions } from '..';
 
-const versions:serviceVersions = {
+const versions: ServiceVersions = {
   admin: '1.0',
   blazegraph: '1.0',
   elasticsearch: '1.0',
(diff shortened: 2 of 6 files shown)


src/client/index.tsx
```diff
@@ -69,9 +69,11 @@ const nexus = createNexusClient({
   uri: preloadedState.config.apiEndpoint,
   links: [setToken],
 });
+
```
dhaneshnm: I will remove these extra spaces in next commit.

src/shared/components/Header/index.tsx
```diff
@@ -2,7 +2,9 @@ import * as React from 'react';
 import { Menu, Dropdown, Icon, Button, Popover } from 'antd';
 import './Header.less';
 import Copy from '../Copy';
-
+import { NexusClient } from '@bbp/nexus-sdk';
```
julienmachon: You cannot make use of the SDK (or any other side effect libraries) in a component. As a global strategy, this is why we have `components` and `containers`

src/shared/components/Header/index.tsx
```diff
@@ -11,6 +13,32 @@ interface InformationContentProps {
   githubIssueURL: string;
 }
 
+const VersionInfo = () => {
+  const preloadedState: RootState = (window as any).__PRELOADED_STATE__;
```
julienmachon: You should not access the `__PRELOADED_STATE__` directly in a component. In fact you should never do this anywhere apart from when you bootstrap your app and you want to dehydrate the global state with the server-side computed state. 

If you need to connect part of your state to a component, use the `react-redux` library and the `connect` higher order function, and finally you need to make sure you do this in a `container` and not a `component`

dhaneshnm:
Screenshot_2019-10-29 A knowledge graph for data-driven science

src/shared/components/Header/__tests__/Header.test.tsx
```diff
@@ -1,17 +1,26 @@
 import * as React from 'react';
 import { shallow, mount } from 'enzyme';
-import Header from '..';
+import Header, {  serviceVersions } from '..';
+
+const versions:serviceVersions = {
```
julienmachon: there are not spaces between your variable and your type. Are you running prettier?

src/shared/components/Header/index.tsx
```diff
@@ -9,8 +9,30 @@ const epflLogo = require('../../EPFL-logo.svg');
 interface InformationContentProps {
   version: string;
   githubIssueURL: string;
+  serviceVersions?: serviceVersions;
 }
 
+export  type serviceVersions = {
```
julienmachon: extra space between `export` and `type`. Also by convention, our types always start with a capital letter
```suggestion
export type ServiceVersions = {
```

src/shared/layouts/MainLayout.tsx
```diff
@@ -37,13 +41,21 @@ const MainLayout: React.FunctionComponent<MainLayoutProps> = ({
...
+  apiEndpoint
 }) => {
   const handleLogout = (e: React.SyntheticEvent) => {
     e.preventDefault();
     localStorage.removeItem('nexus__state');
     userManager && userManager.signoutRedirect();
   };
 
+  
+  const apiBase = new URL(apiEndpoint);
```
julienmachon: clever! :bulb:

src/shared/components/Header/__tests__/Header.test.tsx
dhaneshnm: I have updated the code.

PR #323, opened 2019-10-29
766 update response type in sparql query form

Stafeeva:


Stafeeva:
Implements BlueBrain/nexus#766

PR #324, opened 2019-10-29
Refactor ProjectView 

kenjinp:
implements https://github.com/BlueBrain/nexus/issues/777

Note:
- uses resource List API, therefore the "visit elastic search query from this list" button has been removed.


src/shared/components/DropdownFilter/DropdownItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+const Option = AutoComplete.Option;
+
+export const TypeDropdownItem: React.FunctionComponent<{
+  count: number;
+  key: string;
+  label: string;
+}> = ({ count, key, label }) => {
+  return (
+    // Option does in fact have a label property, the ts warning is a lie.
+    // @ts-ignore (overloaded Options props with label, it works!)
```
julienmachon: Are you sure that you don't need to restart you ts-server or something. I'm pretty sure typescript never lies.

julienmachon: Looks like typescript is right after all

Screenshot from 2019-11-01 15-35-06

src/shared/components/DropdownFilter/DropdownItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <Option key={key} value={key} label={label}>
```
julienmachon: So maybe ?
```suggestion
    <Option key={key} value={key} title={label}>
```

src/shared/store/actions/queryResource.ts
```diff
@@ -94,88 +81,88 @@ const queryResourcesFailedAction: ActionCreator<FailedQueryAction> = (
...
-          schemas,
-          types
-        )
-      );
-    } catch (e) {
-      return dispatch(queryResourcesFailedAction(filterIndex, filterKey, e));
-    }
-  };
-};
+// export const queryResources: ActionCreator<ThunkAction> = (
```
julienmachon: We can assume that this can go then? :laughing:

kenjinp: yes!

src/shared/containers/FileUploadContainer.tsx
```diff
@@ -0,0 +1,73 @@
...
+      projectLabel,
+      {
+        '@id': file.name,
+        file: formData,
+        storage: storageId,
+      },
+      // TODO: fix! https://github.com/BlueBrain/nexus/issues/784
+      // @ts-ignore
+      {
+        extraHeaders: {
```
julienmachon: can you remove this? we can fix it in the SDK. But his extra header isn't fixing anything. Also, not sure why there is a @ts-ignore as extraHeader is a valid option. The issue is created so we can get rid of the comment as well

kenjinp: you got it!

julienmachon:
Good job buddy! :100: / :100:

PR #325, opened 2019-10-31
Fix #778:  use latestRev in updateParams

dhaneshnm:
The history tab will update on revision chnage, when latestRev is part
of updateParams. 
Fixes BlueBrain/nexus#778


julienmachon:
I get an extremely nasty error when I try to update a resource now
Screenshot 2019-10-31 at 14 23 19

PR #327, opened 2019-10-31
✨ FEAT: Use yarn

kenjinp:
it was difficult to upgrade my packages for another task, so I though it's time to keep this repo in line with https://github.com/BlueBrain/nexus-js and convert to yarn. 

- fixes security issues
- upgrades packages
- fixes resulting type errors
- updates snapshots


src/shared/components/FileUpload/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Storage, Project, NexusFile } from '@bbp/nexus-sdk-legacy';
 import { StorageCommon } from '@bbp/nexus-sdk-legacy/lib/Storage/types';
 import { CreateFileOptions } from '@bbp/nexus-sdk-legacy/lib/File/types';
 import { labelOf } from '../../utils';
-import { UploadFile } from 'antd/lib/upload/interface';
+import { UploadFile, RcCustomRequestOptions } from 'antd/lib/upload/interface';
```
julienmachon: What does this have to do with the yarn migration?

src/shared/components/FileUpload/index.tsx
```diff
@@ -74,15 +74,19 @@ const FileUploader: React.FunctionComponent<FileUploaderProps> = ({
...
-  const handleFileUpload = async (customFileRequest: CustomFileRequest) => {
-    try {
-      const options = storageId ? { storage: storageId } : undefined;
-      const nexusFile = await onFileUpload(customFileRequest.file, options);
-      setRecentlyUploadFileList([...recentlyUploadedFileList, nexusFile]);
-      customFileRequest.onSuccess('Successfully uploaded file');
-    } catch (error) {
-      customFileRequest.onError(error);
-    }
+  const handleFileUpload = (customFileRequest: RcCustomRequestOptions) => {
```
julienmachon: Same comment for this one

src/shared/components/FileUpload/index.tsx
kenjinp: The types changed, probably because of version differences

src/shared/components/FileUpload/index.tsx
kenjinp: version differences, I ran `yarn audit fix` its probably why

src/shared/components/FileUpload/index.tsx
julienmachon: gotcha!

PR #328, opened 2019-10-31
782 Show deprecated view

Stafeeva:


Stafeeva:
Implements BlueBrain/nexus#782

Stafeeva:
Yeah I can fix it in this PR. The tricky thing is that when we load an old resource, we get '_deprecated: false' without knowing it was deprecated.

PR #330, opened 2019-11-01
Refactor Image Preview

dhaneshnm:
- The network call to pull an image file for preview is moved to
ResourceView
- ImagePreviewContainer is no longer used in ResourceView

This PR partially implements #BlueBrain/nexus/issues/774. In order to remove Image preview completely, we need to refactor queryItems(@kenjinp is working on it) and InfiniteScroll. For now this only eliminates the use of old sdk in Resource View.


dhaneshnm:
> Should we not wait for this to be merged then? #324
> 
> Not sure I see the value of a partially fixed issue.
> 
> Any thought?

I will wait until @kenjinp  merge his changes. His change will remove the files mentioned in the description, and then the change should be OK.

src/shared/containers/ImagePreviewContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const makePreviewImage = (imageSrc: string) => {
+    const img = new Image();
+    img.src = imageSrc;
+    return <ImagePreviewComponent busy={busy} image={img} />;
+  };
+
+  useAsyncEffect(
+    async isMounted => {
+      if (!isMounted()) {
```
Stafeeva: can it be `!isMounted() || !isFile(resource)`?

kenjinp: good idea

PR #331, opened 2019-11-04
Display Resource metadata in Editor

dhaneshnm:
- Add a toggle to show/hide meta data in editor.
- Fetch and display meta-data, the content is non editable when meta
data is displayed.

Implements, BlueBrain/nexus/issues/779

### DIFF
--- a/src/shared/components/ResourceEditor/index.tsx
+++ b/src/shared/components/ResourceEditor/index.tsx
@@ -9,10 +9,12 @@ export interface ResourceEditorProps {
   rawData: { [key: string]: any };
   onSubmit: (rawData: { [key: string]: any }) => void;
   onFormatChange?(expanded: boolean): void;
+  onMetaDataChange?(expanded: boolean): void;
   editable?: boolean;
   editing?: boolean;
   busy?: boolean;
   expanded?: boolean;
+  metaData?: boolean;
   showExpanded?: boolean;
 }
 
@@ -20,11 +22,13 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
   const {
     rawData,
     onFormatChange,
+    onMetaDataChange,
     onSubmit,
     editable = false,
     busy = false,
     editing = false,
     expanded = false,
+    metaData = false,
     showExpanded = true,
   } = props;
 
@@ -88,6 +92,14 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
               unCheckedChildren="expand"
               checked={expanded}
               onChange={onFormatChange}
+            />
+          )}
+          {!expanded && !isEditing && valid && (
+            <Switch
+              checkedChildren="Meta Data"
+              unCheckedChildren="Show Meta Data"
+              checked={metaData}
+              onChange={onMetaDataChange}
             />
           )}
           {editable && isEditing && valid && (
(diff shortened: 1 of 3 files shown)


dhaneshnm:
Screenshot_2019-11-04 A knowledge graph for data-driven science(2)
Screenshot_2019-11-04 A knowledge graph for data-driven science(1)
Screenshot_2019-11-04 A knowledge graph for data-driven science

src/shared/components/ResourceEditor/index.tsx
```diff
@@ -90,6 +94,14 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
               onChange={onFormatChange}
             />
           )}
+          {!expanded && !isEditing && valid && (
+            <Switch
```
kenjinp: I would put this switch before the expanded toggle, so the expanded toggle doesn't get pushed around if this one isn't moved. 

Also metadata is used as one word in or documentation, so we should do the same here. We can also be more consistent with our copy, using lowercase on all the button toggles here.

kenjinp: Also, there should be a space between the toggles, like so `{" " }`

src/shared/containers/ResourceEditor.tsx
```diff
@@ -23,14 +24,16 @@ const ResourceEditorContainer: React.FunctionComponent<{
...
+  
   const {
     orgLabel,
     projectLabel,
     resourceId,
   } = getResourceLabelsAndIdsFromSelf(self);
   const [{ busy, resource, error }, setResource] = React.useState<{
     busy: boolean;
-    resource: ResourceSource | ExpandedResource | null;
+    resource: ResourceSource | ExpandedResource | Resource |null;
```
Stafeeva: here should be a space after the pipe: `Resource | null`;

src/shared/containers/ResourceEditor.tsx
```diff
@@ -43,6 +46,23 @@ const ResourceEditorContainer: React.FunctionComponent<{
     setExpanded(!expanded);
   };
 
+  const handleMetaDataChange = () => {
+    setShowMetaData(!showMetaData);
+  };
+
+  const getNewResource = async () => {
+    if(expanded) {
```
Stafeeva: please add a space after `if` so it is consistent

src/shared/containers/ResourceEditor.tsx
```diff
@@ -43,6 +46,23 @@ const ResourceEditorContainer: React.FunctionComponent<{
...
+      return  await nexus.Resource.get(orgLabel, projectLabel, resourceId, {
+        rev,
+        format: 'expanded',
+      })
+    } 
+    if(showMetaData) {
```
Stafeeva: same as above

src/shared/containers/ResourceEditor.tsx
```diff
@@ -54,12 +74,8 @@ const ResourceEditorContainer: React.FunctionComponent<{
...
           busy: true,
         });
-        const newResource = expanded
-          ? await nexus.Resource.get(orgLabel, projectLabel, resourceId, {
-              rev,
-              format: 'expanded',
-            })
-          : await nexus.Resource.getSource(orgLabel, projectLabel, resourceId);
+
+        const  newResource =  await getNewResource();
```
Stafeeva: I think there are extra spaces here after `const` and `await` :)

src/shared/components/ResourceEditor/index.tsx
dhaneshnm: Updated the code.

src/shared/containers/ResourceEditor.tsx
dhaneshnm: Ran prettier. They should be good now.

PR #337, opened 2019-11-05
794 remove make public button

Stafeeva:


Stafeeva:
Implements BlueBrain/nexus#794

PR #355, opened 2019-11-11
848 resource graph view layout

kenjinp:
fixes https://github.com/BlueBrain/nexus/issues/848

part of https://github.com/BlueBrain/nexus/issues/820

Screenshot 2019-11-11 at 16 06 20

### DIFF
--- a/src/shared/App.less
+++ b/src/shared/App.less
@@ -12,6 +12,18 @@
     padding: 2em;
     max-width: none;
   }
+  &.resource-view {
+    padding: 0;
+    background-color: #fde7f9;
+    background-image: linear-gradient(315deg, #fde7f9 0%, #e4ebf2 74%);
+    height: 100%;
+    min-height: calc(100vh - 40px);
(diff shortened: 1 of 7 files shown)


src/shared/App.less
```diff
@@ -12,6 +12,18 @@
     padding: 2em;
     max-width: none;
   }
+  &.resource-view {
+    padding: 0;
+    background-color: #fde7f9;
```
Stafeeva: should we use variables for colours?

src/shared/components/Graph/GraphComponent.less
```diff
@@ -0,0 +1,36 @@
...
+    bottom: 0;
+    padding: 1em;
+    margin: 1em;
+  }
+  & > .legend .node {
+    display: inline-block;
+    height: 1em;
+    width: 1em;
+    border-radius: 50%;
+    background-color: #00adee;
```
Stafeeva: variables for colours maybe? :)

PR #356, opened 2019-11-11
create studio list component

julienmachon:
Fixes BlueBrain/nexus#855
Fixes BlueBrain/nexus#856

Just a simple component where studios will be listed
Screenshot from 2019-11-12 14-12-28

You can clone this branch and run the storybook if you wanna test

### DIFF
--- a/src/shared/components/Studio/StudioList.tsx
+++ b/src/shared/components/Studio/StudioList.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react';
+import { Empty, Spin } from 'antd';
 
-import InfiniteSearch from '../List/InfiniteSearch';
 import ListItem from '../List/Item';
 
 import './Studio.less';
-import { Empty, Spin } from 'antd';
 
 type StudioItemProps = {
   id: string;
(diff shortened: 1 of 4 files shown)


src/shared/components/Studio/StudioList.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <div className="studio-list">
+      <h3>Studios</h3>
+      <Spin spinning={busy}>
+        {noStudios && <Empty description="No studios available" />}
+        {!noStudios && (
+          <InfiniteSearch
+            dataLength={studios.length}
+            hasMore={false}
+            onLoadMore={() => {}}
```
Stafeeva: Will it be updated later? Or does InfiniteSearch not work without passing onLoadMore function to it?

julienmachon: good point. Actually, I don't think I need the `InfiniteSearch` component at all... I'll change the code

julienmachon: fixed in e035906f66cd3a0c768c1e50c74e72dc71444813

PR #363, opened 2019-11-12
eneable studio list view and tweak styling

julienmachon:
Fixes https://github.com/BlueBrain/nexus/issues/817

Screenshot 2019-11-12 at 16 51 56

### DIFF
--- a/src/shared/components/Studio/Studio.less
+++ b/src/shared/components/Studio/Studio.less
@@ -3,6 +3,14 @@
 .studio-list {
   .cardlike();
   background-color: @primary-card;
+  width: @default-card-width;
+  min-width: @default-card-width;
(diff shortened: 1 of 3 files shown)


src/shared/components/Studio/Studio.less
```diff
@@ -3,6 +3,14 @@
...
   .cardlike();
   background-color: @primary-card;
+  width: @default-card-width;
+  min-width: @default-card-width;
+  resize: horizontal;
+  overflow: auto;
+  display: flex;
+  flex-direction: column;
+  max-height: 100%;
+  overflow: hidden;
```
Stafeeva: `overflow` is used twice here - is this ok?

PR #370, opened 2019-11-13
854 implement resource preview card

Stafeeva:


Stafeeva:
Screenshot 2019-11-13 at 16 55 27

Stafeeva:
Fixes BlueBrain/nexus#854

kenjinp:
after some discussion the mouse leave should remove the card, because the user might want to copy the id or so from the resource card. Instead, a dismissable X thing should be used in it's place. 

Also, we should have a better styling to display resources. this ResourceCard component is very big for what it does. I've created an issue for that here: https://github.com/BlueBrain/nexus/issues/872

Stafeeva:
Collapsed version:

Screenshot 2019-11-14 at 11 40 33

PR #382, opened 2019-11-15
880 Make origin more visible

Stafeeva:


Stafeeva:
Fixes BlueBrain/nexus#880

PR #384, opened 2019-11-18
888 re center the graph

Stafeeva:


Stafeeva:
Fixes #888

Part of #820

PR #388, opened 2019-11-19
List Dashboard Resources as a Table

dhaneshnm:
Adds container/component to list resources in a Dashboard.

fixes BlueBrain/nexus/issues/833

### DIFF
--- a/src/shared/components/ResultsTable/ResultsTable.tsx
+++ b/src/shared/components/ResultsTable/ResultsTable.tsx
@@ -2,7 +2,9 @@ import * as React from 'react';
 import { Table } from 'antd';
 import './ResultTable.less';
 import moment from 'moment';
-import Search from 'antd/lib/input/Search';
+import { Input } from 'antd';
+
+const { Search } = Input;
 
 const PAGE_SIZE = 10;
 
--- a/src/shared/components/ResultsTable/ResultsTable.tsx
+++ b/src/shared/components/ResultsTable/ResultsTable.tsx
@@ -1,0 +1,113 @@
+import * as React from 'react';
+import { Table } from 'antd';
+import './ResultTable.less';
+import moment from 'moment';
+import Search from 'antd/lib/input/Search';
+
+const PAGE_SIZE = 10;
+
+type ResultTableProps = {
+  headerProperties?: {
+    title: string;
+    dataIndex: string;
+  }[];
+  items: {
+    id: string;
+    [dataIndex: string]: any;
+  }[];
+  pageSize?: number;
+  handleClick: (self: string) => void;
+};
+
+const ResultsTable: React.FunctionComponent<ResultTableProps> = ({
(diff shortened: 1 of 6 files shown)


dhaneshnm:
Screenshot_2019-11-19 A knowledge graph for data-driven science

src/shared/containers/DashboardResultsContainer.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  // build header properties
+  const headerProperties: {
+    title: string;
+    dataIndex: string;
+  }[] =
+    data &&
+    data.head.vars
+      .filter(
+        // we don't want to display total or self url in result table
```
dhaneshnm: This is copied over from studio prototype. I am not sure if this logic makes sense here. ie, Will always have total, self etc in the result?

src/shared/components/ResultsTable/ResultsTable.tsx
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react';
+import { Table } from 'antd';
+import './ResultTable.less';
+import moment from 'moment';
+import Search from 'antd/lib/input/Search';
```
julienmachon: That's a weird import, isn't it?

I would normally do:
```suggestion
import { Input } from 'antd';

const { Search } = Input;
```

src/shared/containers/DashboardResultsContainer.tsx
julienmachon: Yeh, I think we are going to document this. `self` and `total` are normally used for creating links and pagination respectively. We normally don't need to display them in the Data table.

src/shared/containers/DashboardResultsContainer.tsx
```diff
@@ -0,0 +1,104 @@
+import * as React from 'react';
+import { useNexus } from '@bbp/react-nexus';
+import { Spin, Alert } from 'antd';
+import ResulTable from '../components/ResultsTable/ResultsTable';
```
julienmachon: typo. It should be 
```suggestion
import ResultsTable from '../components/ResultsTable/ResultsTable';
```

src/shared/containers/DashboardResultsContainer.tsx
```diff
@@ -0,0 +1,104 @@
...
+}> = props => {
+  const { loading, data, error } = useNexus<any>(
+    nexus =>
+      nexus.View.sparqlQuery(
+        props.orgLabel,
+        props.projectLabel,
+        encodeURIComponent(props.viewId),
+        props.dataQuery
+      ),
+    [props.dataQuery, props.viewId] // only trigger new call if we have a new query and a new view
```
julienmachon: There are a lot of API calls being made when changing the dashboard tab. You can simply remove those 2 args from the array it the API call will only be triggered once
```suggestion
    []
```

PR #397, opened 2019-11-20
890 remove useAsyncEffect

dhaneshnm:
Removes useAsyncEffect from DashboardList, StudioContainer and
WorkspaceList.

Fixes #BlueBrain/nexus/issues/890

### DIFF
--- a/src/shared/containers/DashboardListContainer.tsx
+++ b/src/shared/containers/DashboardListContainer.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { useAsyncEffect } from 'use-async-effect';
 import { Resource } from '@bbp/nexus-sdk';
 import { useNexusContext } from '@bbp/react-nexus';
 import TabList from '../components/Tabs/TabList';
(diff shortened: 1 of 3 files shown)


src/shared/containers/DashboardListContainer.tsx
```diff
@@ -29,22 +28,23 @@ const DashboardList: React.FunctionComponent<DashboardListProps> = ({
...
+          encodeURIComponent(dashboardObject.dashboard)
+        );
+      })
+    )
+      .then(values => {
+        setDashboardResources(values);
+        setSelectedDashboard(values[0]);
+      })
+      .catch(e => {
+        // Fail silently.
```
Stafeeva: we can use andt notification API here to display an error to a user, for example something like this:
```
        notification.error({
          message: `An error occurred`,
          description: error.message,
        });
```

src/shared/containers/StudioContainer.tsx
```diff
@@ -27,17 +26,19 @@ const StudioContainer: React.FunctionComponent<StudioContainerProps> = ({
...
+      .then(value => {
+        if (value['@type'] === 'Studio') {
+          const studioResource: StudioResource = value as StudioResource;
+          setStudioResource(studioResource);
+          const workspaceIds: string[] = studioResource['workspaces'];
+          setWorkspaceIds(workspaceIds);
+        }
+      })
+      .catch(e => {
+        // Fail silently.
```
Stafeeva: Same as above

src/shared/containers/WorkspaceListContainer.tsx
```diff
@@ -23,21 +22,24 @@ const WorkspaceList: React.FunctionComponent<WorkspaceListProps> = ({
...
+          path: workspaceId,
+        });
+      })
+    )
+      .then(values => {
+        setWorkspaces(values);
+        setSelectedWorkspace(values[0] as Resource);
+      })
+      .catch(e => {
+        // Fail silently.
```
Stafeeva: Please :)

src/shared/containers/DashboardListContainer.tsx
dhaneshnm: It was decided earlier that we are going to come back to this later. We have not decided what error to display to the user yet.

PR #400, opened 2019-11-21
Refactor/Simplify graph styles

kenjinp:
- simplify graph style behavior by using only data selectors instead of classes
- make things spiffy by adding transitions
- animate each node's entry with an opacity transition

### DIFF
--- a/src/shared/components/Graph/style.ts
+++ b/src/shared/components/Graph/style.ts
@@ -1,88 +1,113 @@
-const style = [
-    {
-      selector: 'node',
-      style: {
-        height: 16,
-        width: 16,
-      },
+const COLORS = {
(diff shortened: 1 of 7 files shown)


src/shared/components/Graph/style.ts
```diff
@@ -1,88 +1,113 @@
...
-        height: 16,
-        width: 16,
-      },
+const COLORS = {
+  line: '#8a8b8b',
+  blankNode: '#faad14',
+  external: '#00adee',
+  internal: '#ff6666',
+  background: 'white',
+};
```
Stafeeva: love it!

kenjinp: 🦋

PR #401, opened 2019-11-21
Add a cancel button to ResourceEditor

dhaneshnm:
Allow user to cancel the edit and go back to non-redit view.

FIX: #BlueBrain/nexus/issues/799
Screenshot_2019-11-21 A knowledge graph for data-driven science


dhaneshnm:
This will need a different approach to address Kenny's comment.That makes this PR irrelevant.

PR #408, opened 2019-11-22
Bugfix/762 schema links

kenjinp:
fixes https://github.com/BlueBrain/nexus/issues/762


src/shared/containers/SchemaLink.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
+import { useNexusContext } from '@bbp/react-nexus';
+import { Resource } from '@bbp/nexus-sdk';
+
+import { getResourceLabelsAndIdsFromSelf } from '../utils';
+import { useHistory } from 'react-router';
+import SchemaLink from '../components/SchemaLink';
+
+const expandedConstrainedByKey =
```
dhaneshnm: May be this should be in all caps.

PR #411, opened 2019-11-25
read org label and project label correctly + set self to false

julienmachon:
before
Screenshot from 2019-11-25 14-07-08

after
Screenshot from 2019-11-25 14-06-38

### DIFF
--- a/src/shared/views/ACLsView.tsx
+++ b/src/shared/views/ACLsView.tsx
@@ -10,20 +10,12 @@ import ACLsForm from '../components/ACLs/ACLsForm';
 
 interface ACLsViewProps {
   match: any;
-  goToOrg(orgLabel: string): void;
-  goToProject(orgLabel: string, projectLabel: string): void;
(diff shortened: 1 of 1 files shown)


src/shared/views/ACLsView.tsx
```diff
@@ -10,20 +10,12 @@ import ACLsForm from '../components/ACLs/ACLsForm';
 
 interface ACLsViewProps {
   match: any;
```
kenjinp: you can change this to `match: match<{ orgLabel: string; projectLabel: string }>;`

julienmachon: sure, will do. Won't change the fact that they will be no errors if the actual name of the match args are changing somewhere else in the code.

But you gave me an idea

julienmachon: done here 64a5668064173a88bdc29623f61968d714a27f8e

PR #436, opened 2019-12-04
Feature/816 studio master

dhaneshnm:
Merged master.


src/shared/containers/StudioContainer.tsx
```diff
@@ -0,0 +1,73 @@
...
+        }
+      })
+      .catch(e => {
+        // TODO: show a meaningful error to the user.
+      });
+  }, [orgLabel, projectLabel, studioId]);
+  return (
+    <>
+      {studioResource ? (
+        <div className="studio-view">
```
kenjinp: Can this be moved to the studio view as it's presentational? Also see if the `.studio-view` specific css styling in `App.less` can be generalised and removed

PR #444, opened 2019-12-05
929 create studio modal

Stafeeva:
Part of BlueBrain/nexus#902

Fixes BlueBrain/nexus#929

Screenshot 2019-12-04 at 15 47 30

Screenshot 2019-12-04 at 15 47 37

### DIFF
--- a/src/shared/components/Studio/StudioEditorForm.tsx
+++ b/src/shared/components/Studio/StudioEditorForm.tsx
@@ -13,9 +13,7 @@ const StudioEditorForm: React.FC<{
 
     form.validateFields((err, values) => {
       if (!err) {
-        console.log('success', values);
-        
-        saveStudio && saveStudio(values);
+        saveStudio && saveStudio(values.label);
       }
     });
   };
--- a/src/shared/components/Studio/StudioEditorForm.tsx
+++ b/src/shared/components/Studio/StudioEditorForm.tsx
@@ -28,7 +28,7 @@ const StudioEditorForm: React.FC<{
       <Form.Item
         label={
           <span>
-            Label&nbsp;
+            Label{' '}
             <Tooltip title="A name of your studio">
               <Icon type="question-circle-o" />
             </Tooltip>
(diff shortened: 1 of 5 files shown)


dhaneshnm:
Screenshots?

src/shared/containers/CreateStudioContainer.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }).catch(error => {
+      notification.error({
+        message: 'An error occurred',
+        description: error.reason || error.message,
+        duration: 3,
+      });
+    });
+  }
+
+  return (
```
dhaneshnm: Can we move this to a style class/less file?

src/shared/components/Studio/StudioEditorForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+    labelCol: { span: 6 },
+    wrapperCol: { span: 14 },
+  };
+
+  return (
+    <Form {...formItemLayout} onSubmit={handleSubmit}>
+      <Form.Item
+        label={
+          <span>
+            Label&nbsp;
```
dhaneshnm: Can we do {' '} instead?Looks cleaner, IMO

src/shared/components/Studio/StudioEditorForm.tsx
```diff
@@ -0,0 +1,57 @@
...
+    e.preventDefault();
+
+    form.validateFields((err, values) => {
+      if (!err) {
+        saveStudio && saveStudio(values.label);
+      }
+    });
+  };
+
+  const formItemLayout = {
```
dhaneshnm: Can we move this to a style class/less file?

kenjinp: This is an antD layout implementation I don't think it can be ported to a style file. Maybe the best thing is to just remove it?

Stafeeva: we pass it as a js object as prop to antd form

PR #453, opened 2019-12-09
remove studio @type check

kenjinp:
the type can be different depending on project configuration and context, so I suggest to remove this for now


dhaneshnm:
We assume the studio has a workspace property etc and I guess we plan to move to SHACL validated type at some point. Is not checking for type consistent with all this?

kenjinp:
I'll close this because I believe that adding the context to the studio resources and project will solve this issue https://github.com/BlueBrain/nexus-web/pull/454

PR #456, opened 2019-12-10
Fix studio header

Stafeeva:
Fixed Studio Header (it was stretched before)

Screenshot 2019-12-10 at 10 24 09


src/shared/containers/AddWorkspaceContainer.tsx
```diff
@@ -60,6 +60,7 @@ const AddWorkspaceContainer: React.FC<{
           workspaces: updatedWorkspacesList(newWorkspaceId),
           label: studio.label,
           description: studio.description,
+          type: studio['@type'],
```
kenjinp: please destructure the entire studio there, instead of deliberately adding each key

PR #463, opened 2019-12-10
Fix first item in dashboard list rendering

kenjinp:


julienmachon:
Fixes https://github.com/BlueBrain/nexus/issues/936

PR #481, opened 2019-12-18
954 list available plugins

Stafeeva:


Stafeeva:
Fixes BlueBrain/nexus#953

PR #482, opened 2019-12-18
Copy plugin folder to dist

kenjinp:
Copies the `plugins` folder (except for the .gitkeep file) to `dist/public/`


julienmachon:
Well if you do this, you need to remove this code then I guess? https://github.com/BlueBrain/nexus-web/blob/feature/949-nexus-plugin/src/server/index.tsx#L35

PR #487, opened 2020-01-03
961 Studio Docs

kenjinp:
fixes https://github.com/BlueBrain/nexus/issues/961


docs/studio/Dashboards.md
```diff
@@ -0,0 +1,63 @@
...
+Inside any `Studio View`, under a selected `Workspace`, click the `Add Dashboard` button.
+
+![Add a workspace](../assets/add-dashboard-button.png)
+
+..and fill in a form providing the following:
+
+- `Label`: the name of your new `Dashboard` (required field)
+- `Description`: an optional description for your `Dashboard`
+- `Plugins`: an optional list of plugins to add to the details page of a query result
+- `Sparql Query`: a valid [SPARQL](https://www.w3.org/TR/rdf-sparql-query/)query that conforms to the `dataQuery` field in the configuration.
```
julienmachon: The sparql query requires specific headers to be returned in order for studio to work properly such as:
- `?self` for linking to the details page
- `?total` which is a count of the total elements in order to get the pagination working

These should be documented as well and maybe a base query template could be documented too?

kenjinp: I've added documentation for the ?self header, but ?total is not required

PR #506, opened 2020-01-13
Bugfix/975 Fix Workspace Editor

kenjinp:
fixes https://github.com/BlueBrain/nexus/issues/975

## Other changes
- fixes a problem where the current `label` and `description` values would not show if they existed
- changes the label for `Label` from `Name` to `Label` to be more consistent. 

## How to Test

Create and edit a workspace. The edit modal should now show up with the proper values. Before the edit modal would not show.

### DIFF
--- a/src/shared/containers/WorkspaceListContainer.tsx
+++ b/src/shared/containers/WorkspaceListContainer.tsx
@@ -132,7 +132,7 @@ const WorkspaceList: React.FunctionComponent<WorkspaceListProps> = ({
           </div>
         ) : null}
       </TabList>
-      {showEdit && workspaceToEdit ? (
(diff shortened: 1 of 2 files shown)


src/shared/containers/WorkspaceListContainer.tsx
```diff
@@ -132,7 +132,7 @@ const WorkspaceList: React.FunctionComponent<WorkspaceListProps> = ({
           </div>
         ) : null}
       </TabList>
-      {showEdit && workspaceToEdit ? (
+      {showEdit && !!workspaceToEdit ? (
```
dhaneshnm: Do we need to remove !! here?

PR #510, opened 2020-01-15
Bugfix/977 janky studio

kenjinp:
fixes: https://github.com/BlueBrain/nexus/issues/977

resolves jankiness by: 
- moving routing state to query params
- sorts workspace and dashboard results alphabetically, because the ordering is random, which causes jankiness when re-rendering tabs
- removes unnecessary re-rendering by moving components to be direct children of StudioView as a render function (see [here](https://github.com/BlueBrain/nexus-web/compare/Bugfix/977-janky-studio?expand=1#diff-ea6bdbcf23c5771fd39fc5aa667e45a1R38))

adds a `userQueryString` helper hook

### DIFF
--- a/src/shared/containers/DashboardListContainer.tsx
+++ b/src/shared/containers/DashboardListContainer.tsx
@@ -61,11 +61,28 @@ const DashboardList: React.FunctionComponent<DashboardListProps> = ({
           orgLabel,
           projectLabel,
           encodeURIComponent(dashboardObject.dashboard)
-        );
+        ) as Promise<
+          Resource<{
+            label: string;
+            description?: string;
+            dataQuery: string;
+            plugins: string[];
+          }>
+        >;
       })
     )
       .then(values => {
-        setDashboardResources(values);
+        setDashboardResources(
+          values.sort(({ label: a }, { label: b }) => {
+            if (a < b) {
+              return -1;
+            }
+            if (a > b) {
+              return 1;
+            }
+            return 0;
+          })
+        );
         if (
           dashboardId &&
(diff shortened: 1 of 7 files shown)


src/shared/containers/DashboardListContainer.tsx
```diff
@@ -26,11 +26,11 @@ const DashboardList: React.FunctionComponent<DashboardListProps> = ({
   orgLabel,
   projectLabel,
   workspaceId,
-  dashboardId,
   studioResourceId,
   refreshList,
 }) => {
-  const history = useHistory();
+  const [queryParams, setQueryString] = useQueryString();
```
Stafeeva: like it! 👍

src/shared/containers/DashboardResultsContainer.tsx
```diff
@@ -158,11 +136,12 @@ const DashboardResultsContainer: React.FunctionComponent<{
             return binding.self.value.includes(studioResourceId);
           });
         if (selectedResource === undefined && currentResource !== undefined) {
-          selectResource(currentResource.self.value, false);
+          selectResource(currentResource.self.value);
         }
         setItems(tempItems);
       })
       .catch(e => {
+        console.error(e);
```
Stafeeva: do we want to log an error?

kenjinp: probably not!

src/shared/containers/DashboardListContainer.tsx
kenjinp: 🦸🏼‍♂️

PR #523, opened 2020-01-23
984 Add Google Tag Manager

kenjinp:
### How to test

- Start Nexus Web app with GTM_CODE='Your gmt code'

- Open Nexus Web and notice the following popup:

Screenshot 2020-01-24 at 16 12 14

- Click 'Allow' - gtm tag should appear in the page's html. You can also check that preferences are saved to the localStorage.

Screenshot 2020-01-24 at 14 02 04

- To delete gtm, click the 'info' icon on the Header. There you will see the button 'Remove tracking preferences'. Click it, and It will reset your settings in the localStorage, remove the gtm script from the page, and launch same Popup again. If you close it or click 'Don't allow', it will save your preferences to thelocalStorage.

Screenshot 2020-01-24 at 13 47 30

### DIFF
--- a/src/shared/containers/ConsentContainer.tsx
+++ b/src/shared/containers/ConsentContainer.tsx
@@ -1,0 +1,74 @@
+import * as React from 'react';
+import { notification, Button } from 'antd';
+import useLocalStorage from '../hooks/useLocalStorage';
+
+const enableTracking = (trackingCode: string) => {
+  // Do Tracking Stuff
+};
+
+const trackingConsentNotification = (onClose: (accepted: boolean) => void) => {
+  const key = `tracking-consent-notification`;
+  const btn = (
+    <div>
+      <Button
+        size="small"
+        onClick={() => {
+          notification.close(key);
+          onClose(false);
+        }}
+      >
+        Don't Allow
+      </Button>{' '}
+      <Button
+        type="primary"
+        size="small"
+        onClick={() => {
+          notification.close(key);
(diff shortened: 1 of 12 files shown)


src/shared/layouts/MainLayout.tsx
```diff
@@ -55,6 +58,8 @@ const MainLayout: React.FunctionComponent<MainLayoutProps> = ({
     })
   );
 
+  console.log('gtmCode', gtmCode);
```
dhaneshnm: Do we need to log the code in to console?

src/shared/containers/ConsentContainer.tsx
```diff
@@ -0,0 +1,75 @@
...
+        consentToTracking: allow,
+        hasSetPreferences: true,
+      });
+    });
+  }
+
+  if (consent && consent.consentToTracking) {
+    // do google stuff
+    enableTracking(trackingCode);
+    console.log('allowing tracking.....');
```
dhaneshnm: Do we need to log this?

src/shared/layouts/MainLayout.tsx
Stafeeva: It is a draft PR, will be removed

src/shared/containers/ConsentContainer.tsx
Stafeeva: It is a draft PR, will be removed

PR #532, opened 2020-01-28
ResultsTable Sorting

kenjinp:
Fixes https://github.com/BlueBrain/nexus/issues/1012

Implements sorting and (naïve) filtering
filtering options are only enabled if the possible filter values are fewer than 20 and greater than 1

Screenshot 2020-01-29 at 10 33 13

### DIFF
--- a/src/shared/components/ResultsTable/ResultsTable.tsx
+++ b/src/shared/components/ResultsTable/ResultsTable.tsx
@@ -72,13 +72,16 @@ const ResultsTable: React.FunctionComponent<ResultTableProps> = ({
               break;
           }
 
-          const distinctValues = filteredItems.reduce((memo, item) => {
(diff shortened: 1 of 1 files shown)


src/shared/components/ResultsTable/ResultsTable.tsx
```diff
@@ -56,29 +70,59 @@ const ResultsTable: React.FunctionComponent<ResultTableProps> = ({
...
+              if (!memo.includes(value)) {
+                memo.push(value);
+              }
+              return memo;
+            },
+            [] as any[]
+          );
+
+          const filterOptions =
+            distinctValues.length > 1 && distinctValues.length < 20
```
Stafeeva: magic numbers 🤡

PR #534, opened 2020-01-29
1003 studio permissions

dhaneshnm:
Created a studio permission wrapper with AccessControl component and wrapped all studio buttons in it.

### DIFF
--- a/src/shared/components/Tabs/TabElement.tsx
+++ b/src/shared/components/Tabs/TabElement.tsx
@@ -6,9 +6,7 @@ type TabElementProps = {
   label: string;
   description: string;
   onEditClick: (elementId: string) => void;
-  resourcesWritePermissionsWrapper?: (
-    child: React.ReactNode
-  ) => React.ReactNode;
+  permissionsWrapper?: (child: React.ReactNode) => React.ReactNode;
 };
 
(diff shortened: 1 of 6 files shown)


src/shared/components/Tabs/TabElement.tsx
```diff
@@ -32,16 +47,9 @@ const TabElement: React.FunctionComponent<TabElementProps> = ({
...
-          icon="edit"
-          size="small"
-          onClick={e => {
-            onEditClick(id);
-            e.stopPropagation();
-          }}
-          style={buttonStyle}
-        />
+        {studioPermissionsPath
+          ? resourcesWritePermissionsWrapper(editButton, studioPermissionsPath)
```
kenjinp: Lets lift this up into a prop from the parent container because this has logic that doesn't belong in a component (stateful app stuff, specifically the AcessControl using nexusClient)

src/shared/components/Tabs/TabElement.tsx
```diff
@@ -6,17 +6,31 @@ type TabElementProps = {
   label: string;
   description: string;
   onEditClick: (elementId: string) => void;
+  permissionsWrapper?: (child: React.ReactNode) => React.ReactNode;
```
kenjinp: Lets break out the editing behavior as a prop or children

PR #541, opened 2020-01-30
Modal route switch

kenjinp:
Adds modal switch routing like in dribble or pinterest or twitter, for the `ResourceView`

Screenshot 2020-01-30 at 17 00 49

### DIFF
--- a/src/shared/App.tsx
+++ b/src/shared/App.tsx
@@ -6,12 +6,11 @@ import MainLayout from './layouts/MainLayout';
 
 import './App.less';
 import { Modal } from 'antd';
-import ResourceView from './views/ResourceView';
 import ResourceViewContainer from './containers/ResourceViewContainer';
 
 const App: React.FC = () => {
-  let location = useLocation();
-  let history = useHistory();
+  const location = useLocation();
+  const history = useHistory();
 
   // This piece of state is set when one of the
   // gallery links is clicked. The `background` state
--- a/src/shared/App.tsx
+++ b/src/shared/App.tsx
@@ -31,8 +31,9 @@ const App: React.FC = () => {
       </Switch>
 
       {
-        // This is where special routes should go that
-        // That are interacted within a modal to provide outside context
+        // This is where special routes should go
+        // that are placed inside a model
+        // when they background state is provided
       }
       {background && (
         <Route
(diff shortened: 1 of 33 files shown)


src/shared/views/ResourceView.tsx
```diff
@@ -1,367 +1,40 @@
...
-        >
-          {resource && <GraphContainer resource={resource} />}
-        </div>
-      </div>
+      <ResourceViewContainer
+        render={resource => (
+          <div
+            ref={ref}
+            className="graph-wrapper"
+            style={{
```
dhaneshnm: Nit Pick: Do we need these inline styles here? Can we move them to graph-wrapper to create or create a new class?

src/shared/views/ResourceView.tsx
```diff
@@ -1,367 +1,40 @@
...
+              width: '100%',
+              height: '100%',
+              position: 'relative',
+            }}
+          >
+            <div
+              style={{
```
dhaneshnm: Earlier comment about inline styles ..

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -0,0 +1,317 @@
...
+  const handleGoToInternalLink = (link: ResourceLink) => {
+    const { orgLabel, projectLabel } = getOrgAndProjectFromProjectId(
+      (link as IncomingLink)._project
+    );
+    goToResource(orgLabel, projectLabel, encodeURIComponent(link['@id']), {
+      tab: '#links',
+    });
+  };
+
+  useAsyncEffect(async () => {
```
dhaneshnm: Can we use regular useEffect? In the spirit of removing useAsyncEffect from the code base?

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -0,0 +1,317 @@
...
+                            });
+                          }}
+                        >
+                          Revision {rev}
+                        </a>
+                      );
+                    }}
+                  />
+                </TabPane>
+                <TabPane tab="Links" key="#links">
```
dhaneshnm: Can we remove inline styling where possible in the below section?

src/shared/App.tsx
```diff
@@ -1,22 +1,56 @@
...
+    <MainLayout>
+      <Switch location={background || location}>
+        {routes.map(({ path, component: C, ...rest }) => (
+          <Route key={path as string} path={path} component={C} {...rest} />
+        ))}
+        <Route component={NotFound} />
+      </Switch>
+
+      {
+        // This is where special routes should go that
```
Stafeeva: there is probably one extra 'that' in this comment

PR #545, opened 2020-02-03
Rearrange Edit buttons in studio

dhaneshnm:
Change edit and create buttons in studio to match new style.


dhaneshnm:
Screenshot_2020-02-03 A knowledge graph for data-driven science

PR #546, opened 2020-02-03
Bugfix Studio: change desc. from Input to TextArea

kenjinp:
For easier editing of long form text


src/shared/components/DashboardEditor/DashboardConfigEditor.tsx
```diff
@@ -3,6 +3,7 @@ import { Input, Form, Tooltip, Icon, Transfer, Button, Collapse } from 'antd';
 import { WrappedFormUtils } from 'antd/lib/form/Form';
 import { ResourceList } from '@bbp/nexus-sdk';
 import { FormComponentProps } from 'antd/es/form';
+import TextArea from 'antd/lib/input/TextArea';
```
dhaneshnm: can we do import { TextArea } from 'antd' ?

kenjinp: nope :(

PR #550, opened 2020-02-07
Upgrade packages

kenjinp:
In order to remove security warnings in yarn, and to prepare for yarn 2 migration, I've updated our dependencies. 

the process went like this:
- `yarn upgrade`
- fix new type errors created by dependency updates
- `yarn style --write` (prettier was also updated, it seems)
- `yarn test -u` (update snaps)

### DIFF
--- a/src/shared/App.tsx
+++ b/src/shared/App.tsx
@@ -8,6 +8,7 @@ import './App.less';
 import { Modal } from 'antd';
 import ResourceViewContainer from './containers/ResourceViewContainer';
 import StudioResourceView from './views/StudioResourceView';
+import { Location } from 'history';
 
 const App: React.FC = () => {
   const location = useLocation();
@@ -20,7 +21,8 @@ const App: React.FC = () => {
   // use it as the location for the <Switch> so
   // we show the gallery in the background, behind
   // the modal.
-  const background = location.state && location.state.background;
+  const background =
+    location.state && (location.state as { background?: Location }).background;
 
   return (
     <MainLayout>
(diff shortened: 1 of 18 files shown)


src/shared/utils/download.ts
```diff
@@ -1,15 +1,6 @@
...
-    window.navigator.msSaveBlob(blob, filename);
-  } else {
-    const elem = window.document.createElement('a');
-    elem.href = window.URL.createObjectURL(blob);
-    elem.download = filename;
-    document.body.appendChild(elem);
-    elem.click();
-    document.body.removeChild(elem);
-  }
+  window.navigator.msSaveBlob(blob, filename);
```
dhaneshnm: Why did we remove the else block?

src/shared/App.tsx
```diff
@@ -20,7 +21,8 @@ const App: React.FC = () => {
   // use it as the location for the <Switch> so
   // we show the gallery in the background, behind
   // the modal.
-  const background = location.state && location.state.background;
+  const background =
```
dhaneshnm: Are there any changes in Location/history packages? 
location.state.background vs location.state as { background?: Location }).background (?)

kenjinp: The history @types package was changed, yeah :)

src/shared/utils/download.ts
kenjinp: that `window.navigator.msSaveOrOpenBlob` function is guaranteed to exist

PR #562, opened 2020-02-17
Update modals

Stafeeva:
Fixes BlueBrain/nexus#1033

- Changes width to 1100px
- Deletes buttons on Studio Resource View

Screenshot 2020-02-17 at 13 31 17


Stafeeva:
Updated

PR #564, opened 2020-02-17
Fix workspace view select default value

kenjinp:
fixes https://github.com/BlueBrain/nexus/issues/1027

Also changes the success message for editing studio resources from `notification` to `message` because it's more discrete.


Stafeeva:
Looks good! Is it worth changing notifications to messages when we create workspaces/dashboards so it is consistent?

PR #569, opened 2020-02-18
Fix view-statistics type

kenjinp:


src/shared/components/Views/ViewStatisticsProgress.tsx
```diff
@@ -108,8 +108,7 @@ export const ViewStatisticsContainer: React.FunctionComponent<ViewStatisticsCont
     ).subscribe(
       // @ts-ignore
       // TODO fix for v1.3 release
```
Stafeeva: is this comment obsolete? ;)

dhaneshnm: +1

kenjinp: No, it will be after we update the nexus-sdk package to v1.3.1 with the statistics changes (so we dont have to use ts-ignore)

PR #573, opened 2020-02-20
Persistent lists

kenjinp:
Makes resource query lists persistent on your browser

Enables sharing of resource query lists

fixes some bugs from the query list, like if you removed its name.

Changes tsconfig to allow some neat features in typescript like `myMaybeUndefinedObjectPropery?.nowImSafe`

### DIFF
--- a/src/shared/components/ResourceList/index.tsx
+++ b/src/shared/components/ResourceList/index.tsx
@@ -133,7 +133,7 @@ const ResourceListComponent: React.FunctionComponent<{
       <div className="resource-list">
         <h3 className={`header ${busy ? '-fetching' : ''}`}>
           <RenameableItem
-            defaultValue={!!name ? name : 'Unnamed List'}
+            defaultValue={name || 'Unnamed List'}
             onChange={handleUpdate}
             size="small"
           />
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "rootDir": "./",
     "outDir": "./dist",
     "module": "commonjs",
-    "target": "esnext",
+    "target": "es2018",
```
dhaneshnm: Why do we need to change the target?

src/shared/components/ResourceList/index.tsx
```diff
@@ -130,7 +133,7 @@ const ResourceListComponent: React.FunctionComponent<{
       <div className="resource-list">
         <h3 className={`header ${busy ? '-fetching' : ''}`}>
           <RenameableItem
-            defaultValue={name}
+            defaultValue={!!name ? name : 'Unnamed List'}
```
Stafeeva: it can just be `defaultValue={name || 'Unnamed list'}`

kenjinp: yes!

PR #601, opened 2020-03-17
Handle nested objects and arrays

dhaneshnm:
Check for object match with in an array.

### DIFF
--- a/src/shared/utils/__tests__/index.spec.ts
+++ b/src/shared/utils/__tests__/index.spec.ts
@@ -307,7 +307,7 @@ describe('utils functions', () => {
       const plugins: string[] = ['morphology', 'morphology2'];
       const pluginsMap = {
         morphology: {
-          // '@type': ['ReconstructedCell'],
+          '@type': ['ReconstructedCell'],
           distribution: [
             {
               encodingFormat: 'application/swc',
--- a/src/shared/utils/__tests__/index.spec.ts
+++ b/src/shared/utils/__tests__/index.spec.ts
@@ -221,6 +221,7 @@ describe('utils functions', () => {
 
   describe('matchPlugins', () => {
     const plugins: string[] = ['plugin1', 'plugin2'];
+
     const resource: Resource = {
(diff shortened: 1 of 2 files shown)


src/shared/utils/index.ts
```diff
@@ -357,10 +357,20 @@ export const matchPlugins = (
...
     if (Array.isArray(other) && !Array.isArray(value)) {
-      return other.length === 1 && other.indexOf(value) >= 0;
+      return other.length === 1 && isMatch(value, other[0]);
     }
+
+    // return true if any object in value array match any object in
+    // other array.
     if (Array.isArray(other) && Array.isArray(value)) {
-      return difference(other, value).length === 0;
+      for (let i = 0; i < value.length; i += 1) {
```
Stafeeva: it can be just i++ :) classic!

src/shared/utils/__tests__/index.spec.ts
```diff
@@ -301,5 +302,65 @@ describe('utils functions', () => {
       };
       expect(matchPlugins(pluginsMap, plugins, resource)).toEqual([]);
     });
+
+    it('matches a resource with multiple plugins', () => {
+      const plugins: string[] = ['morphology', 'morphology2'];
+      const pluginsMap = {
+        morphology: {
+          // '@type': ['ReconstructedCell'],
```
Stafeeva: is this comment needed here?

src/shared/utils/index.ts
dhaneshnm: Tslint is complaining about ++. https://palantir.github.io/tslint/rules/increment-decrement/
I prefer ++ and -- to be honest.

src/shared/utils/__tests__/index.spec.ts
dhaneshnm: Nope. I have removed it.

src/shared/utils/index.ts
Stafeeva: fair enough :)

PR #607, opened 2020-03-20
Manage seprate index for Resource Array

dhaneshnm:
dashboardResources array and dashboards does not share the same order,
because dashboardResources are sorted and dashboards are unsorted. Hence
a separate variable is needed to track selected dashboard in
dashboardResources.

Fixes BlueBrain/nexus/issues/1121

### DIFF
--- a/src/shared/containers/DashboardListContainer.tsx
+++ b/src/shared/containers/DashboardListContainer.tsx
@@ -189,7 +189,6 @@ const DashboardList: React.FunctionComponent<DashboardListProps> = ({
         tabAction={resourcesWritePermissionsWrapper(tabAction, permissionsPath)}
         editButton={editButtonWrapper}
       >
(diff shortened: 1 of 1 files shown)


src/shared/containers/DashboardListContainer.tsx
```diff
@@ -73,25 +83,28 @@ const DashboardList: React.FunctionComponent<DashboardListProps> = ({
...
-            if (a < b) {
-              return -1;
-            }
-            if (a > b) {
-              return 1;
-            }
-            return 0;
-          })
-        );
+        const sorterValues = values.sort(({ label: a }, { label: b }) => {
```
Stafeeva: should it be `sortedValues`?

dhaneshnm: good catch :)

PR #622, opened 2020-04-29
1133 Global Studio List

Stafeeva:
Fixes BlueBrain/nexus#1133

### This requires some manipulation

- Step 1. Create an `AggregateElasticSeacrhView` somewhere that lists all the default ES views of projects with studios.
Example:
```{
  "@context": "https://bluebrain.github.io/nexus/contexts/view.json",
  "@id": "nxv:studioList",
  "@type": [
    "View",
    "AggregateElasticSearchView"
  ],
  "views": [
    {
      "project": "61025379-8be2-4ef4-87fd-bce39c67e3ba",
      "viewId": "nxv:defaultElasticSearchIndex"
    },
    {
      "project": "aa5dacef-05c6-4278-ad7f-75fa37a79b9d",
      "viewId": "nxv:defaultElasticSearchIndex"
    }
  ]
}
```
 
- Step 2. Pass its org, project and id labels as env: 
`STUDIO_VIEW={orgLabel}/{projectLabel}/{viewId}` when you run the app, for example:

```
STUDIO_VIEW=studios/aggregate-view/nxv:studioList
```
Voila! You have a global studio list

Screenshot 2020-04-29 at 13 54 56


src/shared/components/Studio/ExpandableStudioList.less
```diff
@@ -0,0 +1,38 @@
...
+    background-color: inherit;
+  }
+
+  .studio-title-panel {
+    display: flex;
+    justify-content: space-between;
+  }
+
+  .ant-collapse-arrow {
+    top: 0 !important;
```
dhaneshnm, line 19: Can we find a way to not use !important ? We should use !important as a last resort.

src/shared/components/Studio/ExpandableStudioList.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import './ExpandableStudioList.less';
+
+const ExpandableStudioList: React.FC<{
+  studios: StudioItem[];
+  error: Error | null;
+}> = ({ studios, error }) => {
+  const history = useHistory();
+
+  const makeStudioUri = (studio: StudioItem) => {
```
dhaneshnm, line 16: May be , this belongs to a utility file. With input parameter as a string, rather than studio object. This will make it less specific and re-usable.

dhaneshnm:
General comments,
 Have we tested this for performance? like 1) how long does it take the page to load if we have 10 to 20 studios? 2) Does all the studios load at once, does the UI have any "jumpiness" due to workspaces loading after studio etc. 3) Do we need to think of pagination since we may have more studios than that could fit in one page .

PR #646, opened 2020-07-27
Studio List SubApp

dhaneshnm:
- Move studios in to a sub app and rename it as 'studio legacy'.


dhaneshnm:
Screenshot 2020-07-27 at 15 17 51

PR #679, opened 2020-09-16
Template cards

Stafeeva:
Screenshot 2020-09-16 at 15 10 02


src/subapps/projects/config.tsx
```diff
@@ -0,0 +1,14 @@
+const fusionConfig = {
```
dhaneshnm, line 1: The file extension should be ts, since we don't have any tsx/jsx stuff here.

dhaneshnm (approved):

PR #691, opened 2020-10-07
593 studio refactor

dhaneshnm:
-  Refactor studio to avoid unnecessary re-render. 
- Move Gallery view and related routing to separate components.
- Add an abort in usePlugin to fix memory leak issue. This is happening when you open and close a resource from studio, DashBoardResultsContainer.

Fixes  BlueBrain/nexus/issues/1593


src/shared/hooks/usePlugins.ts
```diff
@@ -52,13 +56,14 @@ export default function usePlugins() {
...
         )
-        .catch(error =>
+        .catch(error => {
           setManifest({
             error,
             loading: false,
             data: null,
-          })
-        );
+          });
```
dhaneshnm, line 64: May be we should make aborting a standard practice across all components that does API calls in UseEffect.

dhaneshnm:
Screenshot 2020-10-08 at 09 17 58

src/shared/views/GalleryView.tsx
```diff
@@ -0,0 +1,92 @@
...
+  // a query param from any app path
+  const self = getUrlParameter('_self');
+  if (self && self !== '') {
+    nexus
+      .httpGet({
+        path: self,
+        headers: { Accept: 'application/json' },
+      })
+      .then((resource: Resource) => {
+        console.log('here ..');
```
Stafeeva, line 33:
```suggestion
```

Stafeeva (approved):
⭐

src/shared/hooks/usePlugins.ts
```diff
@@ -52,13 +56,14 @@ export default function usePlugins() {
...
+          return () => abort.abort();
```
kenjinp, line 65: AbortController requires you pass in a signal for it to work: [https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort)

kenjinp (changes requested):

Stafeeva (approved):

kenjinp (approved):

PR #692, opened 2020-10-08
1651 resources pane

Stafeeva:
Works fine in Chrome and Firefox

Screenshot 2020-10-08 at 10 21 18

### DIFF
--- a/src/subapps/projects/components/ResourcesPane.tsx
+++ b/src/subapps/projects/components/ResourcesPane.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { Collapse } from 'antd';
+import ResizeObserver from 'resize-observer-polyfill';
 
 import ActionButton from './ActionButton';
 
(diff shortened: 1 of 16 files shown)


src/subapps/projects/components/ResourcesPane.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const { Panel } = Collapse;
+
+import './ResourcesPane.less';
+
+const ResourcesPane: React.FC<{}> = ({ children }) => {
+  const [paneWidth, setPaneWidth] = React.useState<number>();
+
+  React.useEffect(() => {
+    const parentDiv = document.getElementsByClassName('activity-view')[0];
```
kenjinp, line 15: get the parent div by using a React.useRef and query ref.current.parent instead of asking for the class name. this way, we will always get the parent div even if we change the class name or something.

kenjinp (changes requested):

dhaneshnm (approved):

kenjinp (approved):

PR #699, opened 2020-10-16
Add code form

Stafeeva:
Fixes BlueBrain/nexus#1656

Screenshot 2020-10-16 at 11 11 03

Screenshot 2020-10-16 at 11 26 09


src/subapps/projects/components/ActionButton.tsx
```diff
@@ -20,8 +20,8 @@ const ActionButton: React.FC<{
   }
 
   const onClickActionButton = (event: any) => {
```
kenjinp, line 22: lets make sure to use the proper typing. You should be able to see the type by hovering over the onClick field of the component.

src/subapps/projects/components/LinkCodeForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+    wrapperCol: { xs: { span: 24 }, sm: { span: 15 } },
+  };
+
+  const columnLayout = {
+    xs: 24,
+    sm: 24,
+    md: 24,
+  };
+
+  const onChangeName = (event: any) => {
```
kenjinp, line 46: same on these...

src/subapps/projects/components/ResourcesPane.tsx
```diff
@@ -34,44 +39,62 @@ const ResourcesPane: React.FC<{}> = ({ children }) => {
...
+        style={{ width: `${paneWidth}px` }}
+      >
+        <Collapse>
+          <Panel
+            header={
+              <div className="resources-pane__header">
+                <span className="resources-pane__pane-name">Resources</span>
+                <ActionButton
+                  title="Link code"
+                  onClick={() => setShowCodeForm(true)}
```
kenjinp, line 64: it's better to reference a variable here, and not write an arrow function, for increased performance.

src/subapps/projects/views/ActivityView.tsx
```diff
@@ -167,6 +171,41 @@ const ActivityView: React.FC = () => {
...
+        if (payload.wasAssociatedWith) {
+          payload.wasAssociatedWith = Array.isArray(payload.wasAssociatedWith)
+            ? [...payload.wasAssociatedWith, { '@id': codeResourceId }]
+            : [payload.wasAssociatedWith, { '@id': codeResourceId }];
+        } else {
+          payload.wasAssociatedWith = { '@id': codeResourceId };
+        }
+
+        activity &&
+          nexus.Resource.update(
```
kenjinp, line 192: Don't nest your .then's and .catch's, but instead return the promise, and add another then() statement on the end of the main one.

kenjinp (changes requested):

src/subapps/projects/components/ResourcesPane.tsx
dhaneshnm: I think not having an arrow function will befit when you have set of elements and we are adding arrow functions to them on a loop. For an element like the above, it looks fine to me.

kenjinp (approved):

dhaneshnm (approved):

PR #700, opened 2020-10-19
Save DashboardId and WrokspaceId in Query String

dhaneshnm:
Save ids in query string so that, create workspace and dashboard works
on default selection.


Stafeeva (commented):
Looks fine, but UI tests broke for some reason :(

Stafeeva (approved):

kenjinp (approved):

PR #710, opened 2020-11-03
Description editor

kenjinp:
Adds a tab to AdminPlugin called `Description`

Adds a markdown editor from `react-mde` 

Markdown can be supplemented with handlebar template syntax to read the resource payload.

You can upload images too! (they will only be visible for public projects for the moment)

Screenshot 2020-11-03 at 18 20 03
Screenshot 2020-11-03 at 19 19 58


adulbrich (approved):

dhaneshnm (approved):

src/shared/hooks/useAsynCall.ts
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
```
dhaneshnm, line 1: We have tow of these files ?

kenjinp: one is a memoized callback, it's used for different purposes. there is `useAsyncCall` and `useAsyncCallback`

dhaneshnm: Ohh, my bad. I suggest we change the name to make the purpose explicit. At present, the names looks very similar.

Stafeeva (approved):

PR #729, opened 2020-11-24
Handle the case when manifest is not present

dhaneshnm:
Internal subapps gets set regardless of whether a manifest is present or
not.


src/subapps/admin/components/Views/ViewStatisticsProgress.tsx
```diff
@@ -15,7 +15,7 @@ export const ViewStatisticsProgress: React.FunctionComponent<ViewStatisticsProgr
   const label =
     percent === 100
       ? `last indexed: ${moment(props.lastIndexed)
-          .startOf()
+          .startOf('days')
```
dhaneshnm, line 18: for some reason, this gives an error with 'days' in local. Not related to subapps.

Stafeeva (approved):

adulbrich (approved):

PR #732, opened 2020-12-01
1816 fetch unlinked activities

Stafeeva:
Resolves BlueBrain/nexus#1816

- Fetches activities with sparql

#### it works with the current data model, and the query should be changed slightly when we update our data model
Screenshot 2020-12-01 at 16 26 59


kenjinp (changes requested):
Please add a tooltip to the Link button and comment out the button.

kenjinp (approved):

PR #744, opened 2020-12-18
1814 Adapt Add new step form

Stafeeva:
Resolves BlueBrain/nexus#1814

- Adapts Step form: Add Step, Step Info - `WorkflowStepFrom`
- Creates `WorkflowStepFormWithActivty` based on old `WorkflowStepFrom`

Screenshot 2021-01-07 at 10 52 38


src/subapps/projects/hooks/useActivitySubClasses.ts
```diff
@@ -0,0 +1,92 @@
...
+      label: string;
+      ['@id']: string;
+    }[]
+  >([]);
+
+  const [datamodelsOrg, datamodelsProject] = dataModelsLocation.split('/');
+
+  const { datamodelsActivityId } = fusionConfig;
+
+  React.useEffect(() => {
```
kenjinp, line 28: if you use `useAsyncCall` here instead, and pass the loading and error props to the component that's using this hook, you can seperate the presentation logic from the fetching logic.

kenjinp (changes requested):
Please move the presentation logic (the  displayError) to the component instead of in the hook.

dhaneshnm (approved):

src/subapps/projects/hooks/useActivitySubClasses.ts
Stafeeva: I haven't use it here as I want to keep values of error and loading in the hook

kenjinp (approved):

PR #745, opened 2021-01-12
1818 adapt new step from notification

Stafeeva:
Screenshot 2021-01-12 at 09 53 40

Screenshot 2021-01-12 at 09 53 56

### DIFF
--- a/src/subapps/projects/components/LinkActivityForm.tsx
+++ b/src/subapps/projects/components/LinkActivityForm.tsx
@@ -3,6 +3,7 @@ import { Col, Row, Button, Select } from 'antd';
 import * as moment from 'moment';
 
 import { getUsername, labelOf } from '../../../shared/utils';
+import TypesIconList from '../../../shared/components/Types/TypesIcon';
 
 import './LinkActivityForm.less';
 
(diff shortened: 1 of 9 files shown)


src/subapps/projects/components/LinkActivityForm.tsx
```diff
@@ -73,7 +74,15 @@ const LinkActivityForm: React.FC<{
             </Col>
             {/* TODO: manage multiple types */}
             <Col>
-              <p>{labelOf(resourceType)}</p>
+              <p className="link-activity-form__types">
+                {resourceType && Array.from(resourceType).length > 0 ? (
+                  <TypesIconList
```
dhaneshnm, line 79: What happens when resourceType is null/undefined etc ? could that happen?

src/subapps/projects/components/NotificationsPopover.tsx
```diff
@@ -31,9 +32,21 @@ const NotififcationsPopover: React.FC<{
         >
           <div className="notifications-popover__main">
             <h4>{activity.name || labelOf(activity.resourceId)}</h4>
-            <p>Created on {moment(activity.createdAt).format('L')}</p>
+            <p className="notifications-popover__types">
+              {activity.resourceType &&
+                Array.from(activity.resourceType).length > 0 && (
```
dhaneshnm, line 37: What happens when resourceType is null/undefined etc ? could that happen?

dhaneshnm (approved):

src/subapps/projects/components/LinkActivityForm.tsx
Stafeeva: We just show nothing if it is the case. We check it here so the app doesn't blow up with error that something is not defined when it tries to render the component but data is still loading.

kenjinp (approved):

PR #760, opened 2021-01-19
Import inputs from search

kenjinp:
Allows importing inputs from Search via the localStorage mechanism.

### DIFF
--- a/src/subapps/projects/containers/InputsContainer.tsx
+++ b/src/subapps/projects/containers/InputsContainer.tsx
@@ -13,7 +13,7 @@ const InputsContainer: React.FC<{
   projectLabel: string;
   stepId: string;
 }> = ({ orgLabel, projectLabel, stepId }) => {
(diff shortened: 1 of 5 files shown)


src/subapps/projects/containers/InputsContainer.tsx
```diff
@@ -1,14 +1,95 @@
...
+
+  React.useEffect(() => {
+    if (!collection || !collection?.ids.length) {
+      return;
+    }
+
+    const key = `${orgLabel}-${projectLabel}-${stepId}`;
+
+    const createInputInWorkflowStep = async () => {
+      const workflowStep = await nexus.Resource.get<{
```
dhaneshnm, line 28: Since we are using async/await, we may also want to put this inside a try catch block and handle the errors.

dhaneshnm (approved):

adulbrich (approved):

PR #765, opened 2021-01-22
Fix linked activities

dhaneshnm:
- Fix query to show the linked activities.
- Fix Sparql Table for an edge case where selected columns is not set.

### DIFF
--- a/src/subapps/projects/containers/ActivityResourcesContainer.tsx
+++ b/src/subapps/projects/containers/ActivityResourcesContainer.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 import { Spin } from 'antd';
 import { useNexusContext } from '@bbp/react-nexus';
-import { isArray } from 'lodash';
 import ResultsTable from '../../../shared/components/SparqlResultsTable';
 import { displayError } from '../components/Notifications';
 import fusionConfig from '../config';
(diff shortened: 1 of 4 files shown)


src/subapps/projects/containers/ActivityResourcesContainer.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Spin } from 'antd';
 import { useNexusContext } from '@bbp/react-nexus';
-
+import { isArray } from 'lodash';
```
kenjinp, line 4: You can use the native `Array.isArray` method instead

src/subapps/projects/containers/ActivityResourcesContainer.tsx
```diff
@@ -16,10 +16,20 @@ const ActivityResourcesContainer: React.FC<{
   linkCodeToActivity: (codeResourceId: string) => void;
 }> = ({ orgLabel, projectLabel, workflowStep, linkCodeToActivity }) => {
   const nexus = useNexusContext();
+  let base;
```
kenjinp, line 19: maybe put this in a function `createWorkflowBaseFromContext` and add it in a lib, then you can create snapshot tests so that we can more easily read what the inputs and outputs look like

kenjinp (approved):
Approved with suggestions

PR #768, opened 2021-01-26
Provide default api mappings for the project

dhaneshnm:
- Provide default api mappings such that, prefixes such as prov, work by
  default.

resolves BlueBrain/nexus/issues/1961.


kenjinp (approved):

adulbrich (commented):
The default API mappings should not include neuroshapes + this should be instance configurable by DKE.

Stafeeva (approved):

PR #769, opened 2021-01-27
Configurable Search Facets

kenjinp:
- allows for users to select which `SearchConfig` to use (locally)
- allows for admins to configure facets in their `SearchConfig`

Known Issue:
- the global search bar won't work without `__original_source` property added in an ES mapping. We might require some design to address this.

### DIFF
--- a/src/subapps/search/views/SearchView.tsx
+++ b/src/subapps/search/views/SearchView.tsx
@@ -83,9 +83,6 @@ const SearchView: React.FC = () => {
     { searchProps, setSearchProps, query },
   ] = useSearchQuery({
     selfURL: preferedSearchConfig?.view,
-    // defaultFacetMap: generateDefaultSearchFilterMap(
-    //   preferedSearchConfig?.facets || []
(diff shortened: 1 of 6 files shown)


dhaneshnm (approved):

src/subapps/search/views/SearchView.tsx
```diff
@@ -68,12 +75,18 @@ const SearchView: React.FC = () => {
...
+    setSearchPreference,
   } = useSearchConfigs();
 
   const [
     searchResponse,
     { searchProps, setSearchProps, query },
-  ] = useSearchQuery(preferedSearchConfig?.view);
+  ] = useSearchQuery({
+    selfURL: preferedSearchConfig?.view,
+    // defaultFacetMap: generateDefaultSearchFilterMap(
```
Stafeeva, line 86: does it need to be here? :)

Stafeeva (approved):

dhaneshnm (approved):

PR #778, opened 2021-02-04
1975 clean up fusion step card

Stafeeva:
Resolves BlueBrain/nexus#1975

Screenshot 2021-02-04 at 10 33 51

### DIFF
--- a/src/subapps/projects/hooks/useStepStatus.ts
+++ b/src/subapps/projects/hooks/useStepStatus.ts
@@ -19,7 +19,7 @@ export const useStepStatus = (orgLabel: string, projectLabel: string) => {
       encodeURIComponent(stepId)
     )
       .then(response => {
-        let originalPayload = response;
+        const originalPayload = response;
 
         return nexus.Resource.update(orgLabel, projectLabel, stepId, rev, {
           ...originalPayload,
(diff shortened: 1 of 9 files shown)


kenjinp (approved):

dhaneshnm (approved):

src/subapps/projects/hooks/useStepStatus.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import fusionConfig from '../config';
+
+export const useStepStatus = (orgLabel: string, projectLabel: string) => {
+  const nexus = useNexusContext();
+  const [error, setError] = React.useState<Error>();
+  const [success, setSuccess] = React.useState<boolean>();
+
+  const updateStatus = async (
+    stepId: string,
```
dhaneshnm, line 12:
```
try {

const source = await nexus.Resource.getSource(
      orgLabel,
      projectLabel,
      encodeURIComponent(stepId)
    );
const updateResponse = await nexus.Resource.update(orgLabel, projectLabel, stepId, rev, {
          ...originalPayload,
          status: newStatus,
          '@type': fusionConfig.workflowStepType,
        });
}
if(updateResonse) {
setSuccess(true)
}
catch (error) {
setError(error)
}
```

dhaneshnm: Some thing like the above would have been better. We should try not to mix the "then" style code with "await".

PR #871, opened 2021-05-28
Serve assets from '/public'

dhaneshnm:


Stafeeva (approved):

src/server/index.tsx
```diff
@@ -52,7 +52,7 @@ app.use(promBundle({ includeMethod: true, metricsPath: `${base}/metrics` }));
 // parse cookies
 app.use(cookieParser());
 // server static assets from the /public directory
-app.use(`${base}/public`, express.static(join(__dirname, 'public')));
+app.use(`/public`, express.static(join(__dirname, 'public')));
```
bogdanromanx, line 55: this seems to go in the other direction; we need to have the client make use of the configured base...

PR #872, opened 2021-05-31
Fix public path

dhaneshnm:
- This will help to compile assets path with base path.


bogdanromanx:
I also noticed this change... but the commit was supposed to solve a CSS image loading issue and was dated on the same day as the 1.4.2 release. This was the commit: https://github.com/BlueBrain/nexus-web/commit/d9942786d6a3c700e38f992487f7269f0bf70c75

Stafeeva (approved):

PR #876, opened 2021-06-07
Fix issue with new activity button

dhaneshnm:
When user click on new button, link the current step as the parent.


src/subapps/projects/containers/ActivitiesLinkingContainer.tsx
```diff
@@ -242,7 +258,7 @@ const ActivitiesLinkingContainer: React.FC<{
           onClickCancel={() => setshowCreateStepForm(false)}
           onSubmit={createNewStep}
           busy={false}
-          parentLabel={''}
+          parentLabel={currentStep ? currentStep['name'] : ''}
```
Stafeeva, line 261: let's use dot notation here, as 'name' has no special characters in it

Stafeeva (approved):

smitfire (approved):

PR #885, opened 2021-06-14
Consolidated notifications

nicwells:
Fixes BlueBrain/nexus#504

If the same notification (notification with same message) is triggered whilst the existing one is still being displayed, the new notification will replace the existing one. Client code should use this notification implementation rather than antd's notification.

PR includes the enhanced notification functionality and replaces calls to antd's notification api in a couple of places but needs to be replaced globally, have opened issue BlueBrain/nexus#2550 to do this once this is approved.

Demo of behaviour using some buttons to test functionality (buttons not in this PR): 
EnhancedNotifications


Stafeeva (approved):
🍦

dhaneshnm (approved):

PR #887, opened 2021-06-16
Include full step path on link activities to workflow step

nicwells:
Fixes BlueBrain/nexus#2045

In Workflow where there are detached activities we are given the option to link them up to steps. There may be substeps belonging to different parent steps that have the same name. Introduced breadcrumbs to parent steps where they exist to disambiguate.

Link_Activity_to_Workflow_step


smitfire (approved):

smitfire:
nice

adulbrich (approved):

PR #889, opened 2021-06-17
Changed defaults values for resources

smitfire:
Changed resource defaults types and added class for links. Need pointer where to implement links though. Looks like the Form.Item component takes any props so could be done there?
https://github.com/BlueBrain/nexus/issues/767
CleanShot 2021-06-17 at 08 47 27@2x

How to test:
head to org -> project -> click 'create resource' button top-mid-right of screen and select accordingly.
example https://host/admin/nick-fusion/newproj

### DIFF
--- a/src/shared/components/ResourceForm/defaultResourcePayloads.ts
+++ b/src/shared/components/ResourceForm/defaultResourcePayloads.ts
@@ -1,66 +1,50 @@
 export const DEFAULT_RESOURCE = {
-  '@context': {}, // optional
   '@id': '', // optional
   '@type': [], // optional
 };
 
 export const DISK_STORAGE = {
-  '@context': {}, // optional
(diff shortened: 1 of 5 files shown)


src/shared/components/ResourceForm/defaultResourcePayloads.ts
```diff
@@ -1,66 +1,50 @@
...
   '@type': 'RemoteDiskStorage',
   default: false,
-  endpoint: '',
-  credentials: '', // optional
-  readPermission: [], // optional
-  writePermission: [], // optional
+  folder: '',
 };
 
 export const S3_STORAGE = {
```
dhaneshnm, line 20: s3 storage is not supported any more. Can we remove this all together (should also remove from the drop down).

PR #892, opened 2021-06-21
Functionality to deprecate workflow step

nicwells:
Fixes BlueBrain/nexus#2426

Deprecating a workflow step can be triggered by the step info modal. Deprecating a workflow step will result in the step, all substeps, and any tables associated with those steps also being deprecated. If the workflow step is the previous step for any steps then it will update those steps to remove that association.

Note: requires a browser refresh currently to refresh workflow page to show updated results.

workflow_delete_step

### DIFF
--- a/src/subapps/projects/utils/index.tsx
+++ b/src/subapps/projects/utils/index.tsx
@@ -121,21 +121,67 @@ export async function fetchChildrenForStep(
     orgLabel,
     projectLabel,
     encodeURIComponent(stepId),
-    'incoming'
+    'incoming',
+    {
+      deprecated: false,
+    }
   );
 
-  const children = await Promise.all(
+  const children = ((await Promise.all(
     links._results
       .filter(link => isParentLink(link))
-      .map(async step => {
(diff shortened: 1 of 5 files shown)


Stafeeva (approved):

src/subapps/projects/utils/index.tsx
```diff
@@ -121,21 +121,67 @@ export async function fetchChildrenForStep(
...
+  const children = ((await Promise.all(
+    uniq.map(async step => {
+      const resource = await nexus.Resource.get(
+        orgLabel,
+        projectLabel,
+        encodeURIComponent(step)
+      );
+      return resource;
+    })
+    // additional filter as ResouceListOptions deprecated option now working
```
dhaneshnm, line 183: I guess it is "Not working".

nicwells: Not working, yes...comment updated in latest commit

dhaneshnm (approved):

Stafeeva (approved):

PR #894, opened 2021-06-21
Update everywhere (almost) to use enhanced notifications

nicwells:
Fixes BlueBrain/nexus#2550

Everywhere except our custom hooks have been updated to use our enhanced notifications rather than antd's notification api directly or the previous helper functions. Updated enhanced notifications implementation to type description as a ReactNode rather than string. Exported helper method based on existing implementation for parsing Nexus error to return a string that can be passed as the description when displaying error notifications.


Stafeeva (approved):
🔥

dhaneshnm (approved):

PR #895, opened 2021-06-21
Fix displaying of polylines when there are multiple previous steps

nicwells:
Fixes BlueBrain/nexus#2496

A step with a single previous would render the connecting polyline correctly however no connecting polyline would be displayed when there are multiple previous steps due to an issue with the polyline id not being set correctly. Updated to set id of polyline correctly when rendering array of prev steps.

2496_workflow_step_links

### Testing
1. Create several workflow steps
2. Click into one of the steps
3. Open the step info and add multiple previous steps
4. Navigate back to the view with the steps
5. Observe the previous steps connecting arrows

### Expected Result
Connecting arrows for each of the previous steps shown on the canvas linked to the step we specified.


dhaneshnm (approved):

Stafeeva (approved):
👨‍🎨

PR #897, opened 2021-06-24
Notification when editing JSON resource added

smitfire:
Added the notification so that users know the file can't be edited. I also added the option component for the actions button but the code was a duplicate of the workflow datatable so I spent a little time exporting it to another component.

However, I don't want to combine features because once exported to a new component the DataTableContainer file should be edited too. Might be worth splitting to a new ticket. 

The code I wanted to export to a new component can be found in DataTableContainer L:1175 - L:243. That way we can just add this option actions component. 

Anyway, though it's super small and the instructions for how test can be found on the issue:  
https://github.com/BlueBrain/nexus/issues/1407

CleanShot 2021-06-24 at 08 36 53@2x

### DIFF
--- a/src/shared/containers/ResourceViewContainer.tsx
+++ b/src/shared/containers/ResourceViewContainer.tsx
@@ -249,6 +249,8 @@ const ResourceViewContainer: React.FunctionComponent<{
     }
   };
 
+  const nonEditableResourceTypes = ['File', 'View'];
+
   React.useEffect(() => {
     setResources();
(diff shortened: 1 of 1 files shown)


dhaneshnm (commented):
Are we checking resource type before showing the alert? Would this show up for all types of resources (not juts files or views)

dhaneshnm (approved):

nicwells (approved):

PR #906, opened 2021-07-09
2565 table viewer

Stafeeva:
Resolves BlueBrain/nexus#2565

### To test:

1. Create resource with @type `Dataset`
2. Create a Nexus File by uploading any `.csv` or `.tsv` file.
3. Link the file to the resource as 'distribution':
```
"distribution": [
    {
      "@type": "DataDownload",
      "contentUrl": "resourceId",
      "name": "filename.csv"
    }
  ],
```

Screenshot 2021-07-09 at 16 09 23

Screenshot 2021-07-09 at 16 00 03

### DIFF
--- a/src/shared/containers/TableViewerContainer.tsx
+++ b/src/shared/containers/TableViewerContainer.tsx
@@ -1,20 +1,21 @@
 import * as React from 'react';
 import { Modal } from 'antd';
-import { Resource, NexusFile } from '@bbp/nexus-sdk';
 import { useNexusContext } from '@bbp/react-nexus';
 import * as csvParser from 'csv-string';
 
 import TableViewer from '../components/TableViewer';
+import useNotification from '../hooks/useNotification';
 
 const TableViewerContainer: React.FC<{
   resourceUrl: string;
   name: string;
   orgLabel: string;
   projectLabel: string;
-}> = ({ resourceUrl, orgLabel, projectLabel, name }) => {
+  onClickClose: () => void;
+}> = ({ resourceUrl, orgLabel, projectLabel, name, onClickClose }) => {
   const nexus = useNexusContext();
   const [tableData, setTableData] = React.useState<any>();
-  const [showTable, setShowTable] = React.useState<boolean>(true);
+  const notification = useNotification();
 
   React.useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/shared/containers/TableViewerContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const TableViewerContainer: React.FC<{
+  resourceUrl: string;
+  name: string;
+  orgLabel: string;
+  projectLabel: string;
+  onClickClose: () => void;
+}> = ({ resourceUrl, orgLabel, projectLabel, name, onClickClose }) => {
+  const nexus = useNexusContext();
+  const [tableData, setTableData] = React.useState<any>();
```
dhaneshnm, line 17: Let us use string[][] instead of any for better type safety. prase method returns string[][]

src/shared/containers/TableViewerContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const notification = useNotification();
+
+  React.useEffect(() => {
+    nexus.File.get(orgLabel, projectLabel, encodeURIComponent(resourceUrl), {
```
dhaneshnm, line 21: Can we use async-await here? You will have to add a method with the async keyword and then call it inside the useEffect. (this is nit pick, you can skip this if you don't feel like doing it :) .

dhaneshnm (approved):

yarn.lock
```diff
@@ -5794,6 +5794,11 @@ csstype@^3.0.2:
   resolved "https://registry.yarnpkg.com/csstype/-/csstype-3.0.8.tgz#d2266a792729fb227cd216fb572f43728e1ad340"
   integrity sha512-jXKhWqXPmlUeoQnF/EhTtTl4C9SnrxSH/jZUih3jmO6lBKr99rP3/+FmrMj4EFpOXzMtXHAZkd3x0E6h6Fgflw==
 
+csv-string@^4.0.1:
```
smitfire, line 5797: i assume we don't need the lock files in PRs.

smitfire (approved):

nicwells (approved):
Nice :)

PR #911, opened 2021-07-28
2609 Modify search bar

Stafeeva:
Resolves BlueBrain/nexus#2609

- Supports project search only
- Uses Dice's Coefficient to find best matches (`string-similarity` js library).

Screenshot 2021-07-28 at 16 32 31


dhaneshnm (approved):

nicwells (approved):
Very nice

PR #915, opened 2021-08-05
1407 updating resources

nicwells:
Fixes https://github.com/BlueBrain/nexus/issues/1407

In Admin, instead of using nexus.Resource.update for everything use the specific endpoint for the resource type when we have a Nexus resource type. This will give us better error messages when there is an error updating a resource which is a Nexus type. Also, some other changes too mainly to improve UX when updating a resource and an error occurs:

* Display error above resource instead of on its own
* Bug fix to editor where parsedValue wasnt always updated if we made a change in the editor and then undid it
* If navigate to resource where doesn't exist, will display the title still and an error message
* Provide link to API documentation when a Nexus error occurs whilst updating resource
* If resource gets updated but there is an error it will reload the resource and display error
* Fix antd warning about using string to specify icon on Resource card

# Example Error
image

### DIFF
--- a/src/shared/containers/ResourceViewContainer.tsx
+++ b/src/shared/containers/ResourceViewContainer.tsx
@@ -268,7 +268,6 @@ const ResourceViewContainer: React.FunctionComponent<{
         busy: false,
       });
     } catch (error) {
-      console.log({ error });
       let errorMessage;
 
       if (error['@type'] === 'AuthorizationFailed') {
@@ -295,7 +294,6 @@ const ResourceViewContainer: React.FunctionComponent<{
       } else {
         errorMessage = error.reason;
       }
-      console.log(errorMessage);
       const jsError = new Error(errorMessage);
 
       setResource({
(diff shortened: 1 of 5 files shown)


dhaneshnm (dismissed):

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -217,6 +268,7 @@ const ResourceViewContainer: React.FunctionComponent<{
         busy: false,
       });
     } catch (error) {
+      console.log({ error });
```
Stafeeva, line 271: forgotten logging :)

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -238,10 +290,12 @@ const ResourceViewContainer: React.FunctionComponent<{
         });
 
         errorMessage = `You don't have the access rights for this resource located in ${orgLabel} / ${projectLabel}.`;
+      } else if (error['@type'] === 'ResourceNotFound') {
+        errorMessage = `Resource '${resourceId}' not found`;
       } else {
         errorMessage = error.reason;
       }
-
+      console.log(errorMessage);
```
Stafeeva, line 298: same

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -261,37 +315,88 @@ const ResourceViewContainer: React.FunctionComponent<{
...
-            )} | ${projectLabel} | ${orgLabel} | Nexus Web`}
-            meta={[
-              {
-                name: 'description',
-                content: getResourceLabel(resource),
-              },
-            ]}
-          />
-        )}
+        {
```
Stafeeva, line 318: I think you don't need these curly braces when you are wrapping stuff in <> </>

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -261,37 +315,88 @@ const ResourceViewContainer: React.FunctionComponent<{
...
+                  <>
+                    <Typography.Paragraph
+                      ellipsis={{ rows: 2, expandable: true }}
+                    >
+                      {error.message}
+                    </Typography.Paragraph>
+                    {error.rejections && (
+                      <Collapse bordered={false} ghost>
+                        <Collapse.Panel key={1} header="More detail...">
+                          {
```
Stafeeva, line 369: Same as above

src/shared/containers/ResourceViewContainer.tsx
nicwells: damn console.log's :)

src/shared/containers/ResourceViewContainer.tsx
nicwells: good spot

Stafeeva (approved):

dhaneshnm (approved):

PR #916, opened 2021-08-10
2623 paginate search results

nicwells:
Fixes https://github.com/BlueBrain/nexus/issues/2623

Pagination of search results. Covers all the user acceptance criteria in issue.

Note: view is a little jumpy on first load and occasionally auto resize of table isn't successful. Looking into this.

### DIFF
--- a/src/subapps/search/views/GlobalSearchView.tsx
+++ b/src/subapps/search/views/GlobalSearchView.tsx
@@ -22,6 +22,10 @@ const GlobalSearchView: React.FC = () => {
   const [queryParams, setQueryString] = useQueryString();
   const { query } = queryParams;
   const [{ ref: wrapperHeightRef }, wrapperDOMProps] = useMeasure();
+  const [
+    { ref: resultTableHeightTestRef },
+    resultTableHeightTestRefDOMProps,
+  ] = useMeasure();
   const [pagination, setPagination] = React.useState({
     numRowsFitOnPage: 0,
     currentPage: 1,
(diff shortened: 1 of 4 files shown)


src/subapps/search/views/GlobalSearchView.tsx
```diff
@@ -177,8 +435,9 @@ const constructQuery = (searchText: string) => {
       prefix_length: 0,
       fields: ['*'],
     })
-    .size(50)
-    .from(0);
+    .rawOption('track_total_hits', true) // accurate total, could cache for performance
+    .size(pageSizeToNotExceedLimit)
```
umbreak, line 439: this will add the from and size to the query payload and not to the URI endpoint, right?

Elasticsearch allows [both](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-from-size.html), to define it on the payload and on the URI (?from={offset}&size={size}).

Currently we enforce `from` and `size` only using query parameters. So If you are going to provide it using the payload, we will have an extra work to do on the backend side to validate those fields (which is fine, but we'll need to plan for it).

nicwells (commented):
@umbreak saw your comment re the from and size params but can't find it now, not sure if I removed somehow whilst trying to reply or if you did but anyway... that's right, it adds to the payload. Unless I'm mistaken, including it in the payload does seem to be working at present. It has also been used elsewhere in Fusion on the views _search endpoint.

An example payload that seems to work on the query endpoint:

```
{
	"from": 10,
	"query": {
		"multi_match": {
			"fields": [
				"*"
			],
			"fuzziness": 5,
			"prefix_length": 0,
			"query": "helo"
		}
	},
	"size": 2,
	"track_total_hits": true
}
```

dhaneshnm (approved):

PR #921, opened 2021-08-19
Fix table height

dhaneshnm:
- Table height works only inline for some reason.

### DIFF
--- a/src/shared/styles/search-tables.less
+++ b/src/shared/styles/search-tables.less
@@ -8,6 +8,10 @@
 @select-blue: rgba(217, 230, 248, 1);
 @select-black: rgba(51, 51, 51, 0.8);
 
+.height-test-wrapper {
+  height: calc(100vh - 82px);
+}
+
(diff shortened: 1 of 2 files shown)


src/shared/styles/search-tables.less
```diff
@@ -8,10 +8,6 @@
 @select-blue: rgba(217, 230, 248, 1);
 @select-black: rgba(51, 51, 51, 0.8);
 
-.height-test-wrapper {
-  height: 'calc(100vh - 82px)';
```
nicwells, line 12: Ah, it's the quotes that's breaking it...should be:

`height: calc(100vh - 82px);`

nicwells (changes requested):
Just need to remove the quotes

src/shared/styles/search-tables.less
dhaneshnm: Makes sense. Will update.

nicwells (approved):

PR #933, opened 2021-08-30
Check for File Resource type

dhaneshnm:
- Check resource type and show 'not editable' when it is a file.
- Show an empty JSON/ '{}' in the editor.

### DIFF
--- a/src/shared/containers/ResourceEditor.tsx
+++ b/src/shared/containers/ResourceEditor.tsx
@@ -35,8 +35,8 @@ const ResourceEditorContainer: React.FunctionComponent<{
 }) => {
   const nexus = useNexusContext();
   const notification = useNotification();
-  const [expanded, setExpanded] = React.useState(defaultEditable);
-  const [editable, setEditable] = React.useState(defaultExpanded);
+  const [expanded, setExpanded] = React.useState(defaultExpanded);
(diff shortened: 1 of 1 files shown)


src/shared/containers/ResourceEditor.tsx
```diff
@@ -30,7 +35,8 @@ const ResourceEditorContainer: React.FunctionComponent<{
 }) => {
   const nexus = useNexusContext();
   const notification = useNotification();
-  const [expanded, setExpanded] = React.useState(defaultExpanded);
+  const [expanded, setExpanded] = React.useState(defaultEditable);
+  const [editable, setEditable] = React.useState(defaultExpanded);
```
nicwells, line 39: should be the other way around no?

```
const [expanded, setExpanded] = React.useState(defaultExpanded);
const [editable, setEditable] = React.useState(defaultEditable);
```

nicwells (commented):
Just one little thing

src/shared/containers/ResourceEditor.tsx
dhaneshnm: Yes. Will change.

Stafeeva (approved):

nicwells (approved):

PR #935, opened 2021-09-02
Fixed table layout, refactor field visibility

nicwells:
Column width is now fixed to 200px although will increase to fill table if there are fewer columns than is enough to fill width. If there are more columns than fit then the table will expand horizontally.

Refactored field visibility to use a reducer hook.

### DIFF
--- a/src/subapps/search/hooks/useGlobalSearch.tsx
+++ b/src/subapps/search/hooks/useGlobalSearch.tsx
@@ -65,6 +65,70 @@ function filterReducer(
   }
 }
 
+function fieldVisibilityReducer(
+  state: FieldsVisibilityState,
+  action: fieldVisibilityActionType
+): FieldsVisibilityState {
+  switch (action.type) {
+    case 'initialize':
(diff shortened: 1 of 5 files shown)


Stafeeva (approved):

src/subapps/search/hooks/useGlobalSearch.tsx
```diff
@@ -65,6 +65,70 @@ function filterReducer(
   }
 }
 
+function fieldVisibilityReducer(
```
dhaneshnm, line 68: local storage change is technically a side effect like dom updates, network calls etc. Instead of doing this in reducer, we can move this to a useEffect that just handles all local storage updates.

useffect(() => {
      localStorage.setItem(
        'search-field-visibility',
        JSON.stringify(FieldsVisibilityState) 
      ); // Something of this sort.
}, [FieldsVisibilityState])

We don't need to do this in this PR, can create a tech debt issue and address this late.

dhaneshnm (approved):

PR #985, opened 2021-11-02
2807 admin layout tab update

smitfire:
Fixes #

## Description

Updated the project admin view to include a few more options. Permissions are now combined with settings (previously only accessible from projects list view).
Statistics and Querying is also now in a separate tab and does not follow an external link.
Studios, Workflow links are embedded in the tab layout but go to the original page via new browser tab.
Create upload moved to new tab, no changes to component.

## How has this been tested?

Dev and local environment using the compositepoc org and all of the projects within it. 
I have removed the banners from the query, stats, settings components because it is now redundant.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have added screenshots (if applicable), in the comment section.

### DIFF
--- a/src/shared/styles/project-banner.less
+++ b/src/shared/styles/project-banner.less
@@ -1,7 +1,6 @@
 @import '../lib.less';
 
 .project-banner {
-  // background-color: fade(@header-background-color, 2.5%);
   padding: 0 @default-pad;
   display: flex;
   justify-content: space-between;
--- a/src/shared/styles/project-banner.less
+++ b/src/shared/styles/project-banner.less
@@ -1,7 +1,7 @@
 @import '../lib.less';
 
 .project-banner {
-  background-color: fade(@header-background-color, 2.5%);
+  // background-color: fade(@header-background-color, 2.5%);
   padding: 0 @default-pad;
   display: flex;
   justify-content: space-between;
--- a/src/subapps/admin/components/Projects/QueryEditor.tsx
+++ b/src/subapps/admin/components/Projects/QueryEditor.tsx
@@ -1,0 +1,46 @@
+import * as React from 'react';
+import { Link } from 'react-router-dom';
+import {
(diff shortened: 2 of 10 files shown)


dhaneshnm (approved):

src/shared/styles/project-banner.less
```diff
@@ -1,7 +1,7 @@
 @import '../lib.less';
 
 .project-banner {
-  background-color: fade(@header-background-color, 2.5%);
+  // background-color: fade(@header-background-color, 2.5%);
```
nicwells, line 4: remove rather than comment out

src/subapps/admin/components/Projects/QueryEditor.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <div className="query-editor">
+      <h3>Query Browser</h3>
+      <p>
+        View resources in your project using pre-defined query-helper lists.
+      </p>
+      <div className="project-menu__controls">
+        <Tabs defaultActiveKey="browse" tabPosition="left">
+          <TabPane tab="SparQL" key="browse">
+            <div style={{ flexGrow: 1 }}>
```
nicwells, line 21: styles in separate less file? If copy/paste from elsewhere then can leave

nicwells (changes requested):
There are a few issues picked up by Deep Scan. Also checking against acceptance criteria:

* Deprecate project button not working so was not able to test acceptance criteria related to this
* Settings button still showing in projects listing
* Clicking on studios/workflows opens in new tab/window but then going back to previous view shows tab with no contents. Possible not to switch to the empty tab when clicked?

dhaneshnm (approved):

smitfire:
@nicwells removed settings button, likewise for deprecated. made the change for clicking studio or workflows

PR #1013, opened 2022-01-03
added numerics and conditions to call numerics component to search

smitfire:
Fixes #

Screenshot 2022-01-03 at 13 41 56
Screenshot 2022-01-03 at 13 42 02
Screenshot 2022-01-03 at 13 42 05

## Description

Added the numerics filter logic but since all the fields with numerics are actually returning strings it is not convenient to test. I have added some preliminary logic to convert the data to actual integers.

The logic for the min, max for the slider range is there too. As well as the show missing checkbox condition but the logic connecting it to the missing query is not added.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] My change requires a change to the documentation.
- [x ] I have added screenshots (if applicable), in the comment section.


dhaneshnm (approved):

nicwells (commented):
* Show missing values doesn't appear to be working for me, seems to have no effect and doesn't disable other controls (related to next point)
* Clicking on column dropdown arrow results in filter being applied immediately and missing values being removed from results
* between filter looks a little strange and seems to be repeated (see screenshot below)
Screenshot 2022-01-06 at 10 25 15

* Couple of styling comments: when numbers is over a certain size it doesn't fit in text box, right side of slider is a bit close to max text box, would be nice to format number of missing values to include comma separator
Screenshot 2022-01-06 at 10 29 00


* Crashes for me when opening the filter for Bouton Density when pointing to dev - not sure if this is a data thing or not

src/subapps/search/utils/index.ts
```diff
@@ -47,6 +54,28 @@ const missingFilterValueAdder = (filterTerm: string) => {
     missing.notFilter('exists', filterTerm);
 };
 
+export const constructNumberFilter = (
+  body: bodybuilder.Bodybuilder,
+  filters: string[],
+  filterType: string,
+  filterTerm: string
+) => {
+  console.log(filters);
```
nicwells, line 63: remove this

dhaneshnm (approved):

nicwells (commented):
Looks good :) Just the console.log to remove

PR #1175, opened 2022-11-01
Using _self and @id to reference report assets

smitfire:
Fixes #https://github.com/BlueBrain/nexus/issues/3548

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/containers/AnalysisPlugin/AnalysisPluginContainer.tsx
+++ b/src/shared/containers/AnalysisPlugin/AnalysisPluginContainer.tsx
@@ -49,6 +49,7 @@ async function fetchImageObjectUrl(
   const blob = new Blob([rawData as string], {
     type: encodingFormat,
   });
+
   return URL.createObjectURL(blob);
 }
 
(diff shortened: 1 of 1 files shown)


src/shared/containers/AnalysisPlugin/AnalysisPluginContainer.tsx
```diff
@@ -405,6 +408,8 @@ const AnalysisPluginContainer = ({
       scripts?: ReportGeneration[];
     }) => {
       const unsavedAssetsToAddToDistribution = unsavedAssets.map(a => {
+        console.log('unsaved assets to add to distribution');
```
imsdu, line 411: We may want to delete these

src/shared/containers/AnalysisPlugin/AnalysisPluginContainer.tsx
```diff
@@ -284,10 +285,11 @@ const AnalysisPluginContainer = ({
           const imageId = asset.filePath.substring(
             asset.filePath.lastIndexOf('/') + 1
           );
-          const imgResource = (await nexus.Resource.get(
+
+          const fileResource = (await nexus.File.get(
             orgLabel,
             projectLabel,
-            encodeURIComponent(asset.filePath)
+            encodeURIComponent(imageId)
```
imsdu, line 292: `fileId` instead of `imageId`

imsdu (changes requested):

imsdu (approved):

src/shared/containers/AnalysisPlugin/AnalysisPluginContainer.tsx
smitfire: imageId is the uuid renaming it is not needed. it is also used elsewhere

PR #1213, opened 2023-04-21
3345 // Changes to table configuration should save correctly

Dinika:
If the user updated a column config in the column, the next time they open the "Edit Table" dialog, the changes they made previously should still be preserved and shown.

Fixes #[3345](https://app.zenhub.com/workspaces/blue-brain-nexus-60ace035034cae00105c9307/issues/gh/bluebrain/nexus/3345)

## How to test this PR

1. Navigate to any studio that you are allowed to edit ({baseUrl}/nexus/studios)
    a. Create a workspace if there aren't any
2. Open the add dashboard dialog.
    a. Give any name to dashboard. Select sparql as the view from the dropdown. ("https://bluebrain.github.io/nexus/vocabulary/defaultSparqlIndex")
    b. **Test** that the "Column Configuration" section is there with 3 checkboxes per column. 
    Screenshot 2023-04-27 at 15 54 49
    
3. Check a couple of these checkboxes. Save the dashboard. 
4. Click edit dashboard button. 
    a. **Test** that the previously checked checkboxes are still checked
5. Reload page, open edit dashboard dialog
    a. **Test** that the previously checked checkboxes are still checked

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/components/EditTableForm.tsx
+++ b/src/shared/components/EditTableForm.tsx
@@ -1,40 +1,41 @@
-import * as React from 'react';
+import { View } from '@bbp/nexus-sdk';
 import { useNexusContext } from '@bbp/react-nexus';
 import {
+  Button,
+  Checkbox,
+  Col,
   Form,
   Input,
-  Button,
-  Spin,
-  Checkbox,
   Row,
-  Col,
   Select,
+  Spin,
   Tooltip,
 } from 'antd';
-import { Controlled as CodeMirror } from 'react-codemirror2';
 import 'codemirror/addon/display/placeholder';
-import 'codemirror/mode/sparql/sparql';
-import 'codemirror/mode/javascript/javascript';
+import 'codemirror/addon/fold/brace-fold';
 import 'codemirror/addon/fold/foldcode';
 import 'codemirror/addon/fold/foldgutter';
-import 'codemirror/addon/fold/brace-fold';
 import 'codemirror/lib/codemirror.css';
+import 'codemirror/mode/javascript/javascript';
+import 'codemirror/mode/sparql/sparql';
 import 'codemirror/theme/base16-light.css';
-import { View } from '@bbp/nexus-sdk';
+import * as React from 'react';
+import { Controlled as CodeMirror } from 'react-codemirror2';
 import { useQuery } from 'react-query';
-import ColumnConfig from './ColumnConfig';
+import { FUSION_TABLE_CONTEXT } from '../../subapps/projects/fusionContext';
 import {
-  queryES,
-  parseESResults,
-  querySparql,
-} from '../hooks/useAccessDataForTable';
-import './EditTableForm.less';
-import {
-  TableResource,
   TableColumn,
+  TableResource,
   UnsavedTableResource,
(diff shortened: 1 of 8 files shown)


src/shared/components/ColumnFilterDropDown.tsx
```diff
@@ -0,0 +1,44 @@
+import { Button, Input, Space } from 'antd';
+import { FilterDropdownProps } from 'antd/lib/table/interface';
+import React from 'react';
+
+export const ColumnFilterDropDown: (
+  props: FilterDropdownProps & { label: string }
```
shinyhappydan, line 6: What relationship does this component have to antd? It looks to me like the fields are the same by coincidence rather than being exactly the same thing?

src/shared/components/EditTableForm.tsx
```diff
@@ -283,7 +284,9 @@ const EditTableForm: React.FC<{
     },
     {
       onSuccess: data => {
-        setConfiguration(data);
+        if (isNil(configuration)) {
+          setConfiguration(data);
+        }
```
shinyhappydan, line 289: Just curious why this is better than `if(!configuration)`

src/shared/components/ColumnFilterDropDown.tsx
Dinika: Thanks for the review :)

To begin with, I have now removed these changes (apologies, the PR was in DRAFT state. I shouldn't have added reviewers).

That said, if we did upgrade antd the component `ColumnFilterDropDown` would not break (unless antd rename or remove `FilterDropdownProps` interface). Antd allows us to pass any custom react component as a value to its `filterDropdown` prop. As long as our custom component accepts the same (or an extension of) shape of props as `FilterDropdownProps` or its extension, it is correct from ts point of view.

src/shared/components/EditTableForm.tsx
Dinika: O, just cuz (imho) `isNil` is more deliberate/explicit about what it checks (it checks for `null` and `undefined`) while simple falsy checks can return false for a number of other values.
In any case, this line has now been remoed.

src/subapps/studioLegacy/components/WorkspaceEditorForm.tsx
```diff
@@ -59,7 +59,10 @@ const WorkspaceEditorForm: React.FC<{
           },
         ]}
       >
-        <Input className="ui-workspace-label-input" />
+        <Input
+          className="ui-workspace-label-input"
+          data-testid="workspace-label"
```
smitfire, line 64: small recommendation. consider 'data-test-id'.

Dinika: Interestingly, when I change this to `data-test-id` then the `findByTestId` queries in cypress and testing-library stop working. Looks like [they need](https://testing-library.com/docs/queries/bytestid/) the attribute to be spelled without the two dashes.

smitfire: ok if that's the case then go with 'data-testId'

smitfire (approved):

PR #1220, opened 2023-05-10
Feature 3224 // studio data download

Dinika:
Fixes #[3224](https://app.zenhub.com/workspaces/blue-brain-nexus-60ace035034cae00105c9307/issues/gh/bluebrain/nexus/3224)
Screenshot 2023-05-11 at 19 24 48

## Description

Enable users to download metadata and files associated with rows in studios table.

## How to test
1. Open a dashboard in studios
2. Select a row to download
3. Click on download button in datapanel -> A tar file should be downloaded.

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/src/shared/hooks/useAccessDataForTable.tsx
+++ b/src/shared/hooks/useAccessDataForTable.tsx
@@ -1,19 +1,47 @@
+import { FilterFilled } from '@ant-design/icons';
 import { NexusClient, Resource, SparqlView, View } from '@bbp/nexus-sdk';
 import { useNexusContext } from '@bbp/react-nexus';
+import * as Sentry from '@sentry/browser';
 import { ColumnType } from 'antd/lib/table/interface';
 import * as bodybuilder from 'bodybuilder';
 import json2csv, { Parser } from 'json2csv';
-import { isNil, isString, pick, toNumber } from 'lodash';
+import {
+  has,
+  isArray,
+  isNil,
+  isString,
+  pick,
+  sumBy,
+  toNumber,
+  uniq,
+  uniqBy,
+} from 'lodash';
 import * as React from 'react';
 import { useQuery } from 'react-query';
+import {
+  StudioTableRow,
+  TableError,
+  getStudioRowKey,
+} from '../../shared/containers/DataTableContainer';
+import {
+  MAX_DATA_SELECTED_SIZE__IN_BYTES,
+  MAX_LOCAL_STORAGE_ALLOWED_SIZE,
+  TDataSource,
+  TResourceTableData,
+  getLocalStorageSize,
+  notifyTotalSizeExeeced,
+} from '../../shared/molecules/MyDataTable/MyDataTable';
+import {
+  DATA_PANEL_STORAGE,
+  DATA_PANEL_STORAGE_EVENT,
+} from '../../shared/organisms/DataPanel/DataPanel';
+import { fileExtensionFromResourceEncoding } from '../../utils/contentTypes';
 import { Projection } from '../components/EditTableForm';
 import { download } from '../utils/download';
 import { isNumeric, parseJsonMaybe } from '../utils/index';
 import { addColumnsForES, rowRender } from '../utils/parseESResults';
 import { sparqlQueryExecutor } from '../utils/querySparqlView';
 import { CartContext } from './useDataCart';
-import { FilterFilled } from '@ant-design/icons';
-import { TableError } from 'shared/containers/DataTableContainer';
 
 export const EXPORT_CSV_FILENAME = 'nexus-query-result.csv';
 export const CSV_MEDIATYPE = 'text/csv';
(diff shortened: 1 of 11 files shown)


src/shared/App.tsx
bilalesi, line 1: the data panel will be displayed at all pages of the webapp when there is data in it,
please merge develop to get the last update,

src/shared/organisms/DataPanel/DataPanel.tsx
```diff
@@ -603,7 +649,6 @@ const DataPanel: React.FC<Props> = ({}) => {
               className="my-data-panel-table"
               rowClassName="my-data-panel-table-row"
               pagination={false}
-              scroll={{ y: 400 }}
```
bilalesi, line 606: why remove the scroll ?

src/utils/contentTypes.ts
bilalesi, line 1: may you explain why you add the whole file extensions, since in our case the user has a limited type of data, and may not be in this list as (`SWC`, `OBJ`), so we can stick to the file extensions extracted from the file name or from the encoding format

src/shared/hooks/useAccessDataForTable.tsx
```diff
@@ -360,6 +386,81 @@ const accessData = async (
...
+                `${resourceName}.${fileExtensionFromResourceEncoding(
+                  distItem.encodingFormat,
+                  'txt'
+                )}`;
+              return itemLabel;
+            })
+          : resource.distribution?.label ??
+            `${resourceName}.${fileExtensionFromResourceEncoding(
+              resource.distribution?.encodingFormat,
+              'txt'
```
bilalesi, line 431: why a default file extension as `txt`, this may damage the file if it's not a txt, 
I prefer to let it as it is (no extension) and the user can decide which app can open the file (The user know better the data he/she has)

src/shared/hooks/useAccessDataForTable.tsx
```diff
@@ -360,6 +386,81 @@ const accessData = async (
...
+            )}`,
+        hasDistribution: has(resource, 'distribution'),
+      },
+    };
+  } catch (err) {
+    Sentry.captureException(err, {
+      extra: {
```
bilalesi, line 438: why do you think about adding the user who request the data download,
it will help if we need more investigation
```
User {
    id?: string;
    ip_address?: string;
    email?: string;
    username?: string;
}
```

bilalesi: you can get the user object from the global state

src/shared/organisms/DataPanel/DataPanel.tsx
```diff
@@ -99,8 +98,19 @@ function findIndexes(arr: any[], predicate: any) {
   });
   return indexes;
 }
-function removeIndeces(arr: any[], indicesToRemove: number[]) {
-  return arr.filter((_, index) => !indicesToRemove.includes(index));
+
+function getPathForChildResource(resource: any, parent: ResourceObscured) {
+  const extension = resource._filename?.split('.').pop() ?? 'txt';
```
bilalesi, line 103: same here, are we sure that the data type is always `txt` in studios

src/shared/organisms/DataPanel/DataPanel.tsx
Dinika: My bad. Thanks for pointing out. FIxed it.

src/utils/contentTypes.ts
Dinika: Well, in studios, the `encodingFormat` I get back form server are these (just a few example of actual ones I got from server):

```
text/plain ; UTF-256
text/turtle
application/ld+json
text/csv
text/json
```
I do not get an extension in label for filename. This seemed like the simplest way to extract extension from mime type.

src/shared/hooks/useAccessDataForTable.tsx
Dinika: agreed. I don't add a default extension now.

src/shared/organisms/DataPanel/DataPanel.tsx
Dinika: got it. removed.

src/shared/hooks/useAccessDataForTable.tsx
Dinika: I actually do like the idea of adding more information to our logs, but in this instance I don't find user's information useful in debugging. Here I am only logging an error when a resource could not be serialized. Since this issue is not user specific - i.e. any user would get the same error if they tried to add this resource to cart, I don't think logging user details is pertinent here.

Also, since it also contains username and email, I think it's also not secure to log it without anonymizing. 

I guess, what we should do though at some point is add more generic information (like route, perhaps project or org details) to all our sentry logs. In many cases, user information could also be added after anonymizing it though.

src/utils/contentTypes.ts
Dinika: Just for reference this is what I get back from the server in the `distribution` object:

```json
[
    {
        "@type": "DataDownload",
        "contentSize": {
            "unitCode": "bytes",
            "value": 723725
        },
        "contentUrl": "https://staging.nise.bbp.epfl.ch/nexus/v1/files/bbp/nmc/00c66a76-ef47-4366-acda-7adb1884b641",
        "digest": {
            "algorithm": "SHA-256",
            "value": "f1dcfa60ca11670739ae8deebbaf580d876beba9107ac422b3a3ad1f907b9bed"
        },
        "encodingFormat": "text/plain; charset=UTF-8",
        "name": "file"
    },
    {
        "@type": "DataDownload",
        "contentUrl": "http://microcircuits.epfl.ch/#/article/article_3_mph",
        "repository": {
            "@id": "http://microcircuits.epfl.ch/#/article/article_3_mph"
        }
    }
]
```

bilalesi (dismissed):

bilalesi (approved):

PR #1229, opened 2023-05-18
Hotfix/my data pagination

bilalesi:
Fixes: pagination size changer in my-data page
Screenshot 2023-05-18 at 10 55 53

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/molecules/MyDataTable/MyDataTable.tsx
+++ b/src/shared/molecules/MyDataTable/MyDataTable.tsx
@@ -1,4 +1,5 @@
 import React, { Fragment, useMemo, useReducer, useEffect } from 'react';
+import * as moment from 'moment';
 import { Button, Table, Tag, Tooltip, notification } from 'antd';
 import { Link, useHistory, useLocation } from 'react-router-dom';
 import { PaginatedList } from '@bbp/nexus-sdk';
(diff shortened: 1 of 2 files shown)


src/shared/molecules/MyDataTable/MyDataTable.tsx
```diff
@@ -254,15 +255,17 @@ const MyDataTable: React.FC<TProps> = ({
         title: 'updated date',
         dataIndex: 'updatedAt',
         width: 140,
-        sorter: false,
+        sorter: (a, b) =>
+          moment(a.updatedAt).unix() - moment(b.updatedAt).unix(),
```
Dinika, line 259: Since `moment` has reached it's end of life, I would strongly discourage it's further use in the project, so we can gradually move away from it.

Fortunately for simple use cases, such as these the native `Date` api is already pretty good:

I think, for us this should suffice:
```javascript
new Date(a.updatedAt) - new Date (b.updatedAt)
```

if you still prefer to use ms since epoch, we can use [getTime()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime):
```javascript
new Date(a.updatedAt).getTime() - new Date (b.updatedAt).getTime()
```

Note: we might want to check for possible undefined values.

Dinika (changes requested):

PR #1239, opened 2023-05-24
update: use baseURl instead of window location

bilalesi:
- default Redirect is home page so to avoid double slash '//' in the route (may it be temporary solution)
- use baseURl instead of window location to get the real location

## Types of changes

## Checklist:


Dinika (approved):
Lgtm. The pipeline is failing. Once it's fixed, we can merge.

PR #1248, opened 2023-05-31
3896 // Improve data download when resources have multiple distribution items

Dinika:
Fixes #3896

## Description
This PR improves the experience for downloading resources that have multiple distributions attached to it. 

This is how the data-panel would look before when a user selected a row with multiple distributions in studios:
Screenshot from 2023-05-31 16-22-21
(The data panel only shows one file extension)

This is how the data panel will look now:
Screenshot from 2023-05-31 16-22-41
(The data panel has extensions for all the files attached to this row)

## How has this been tested?

1. I've tested this for resources with distribution arrays and with no distributions (both manually and with some unit tests)
2. I've also tested that the download continues to work for my-data and search 

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.


src/shared/organisms/DataPanel/DataPanel.tsx
```diff
@@ -467,7 +486,7 @@ const DataPanel: React.FC<Props> = ({}) => {
                 : 'Resource',
             resourceId: resource.id,
             project: `${parsedSelf.org}/${parsedSelf.project}`,
-            path: `/${parsedSelf.project}/${pathId}${
+            path: `/${parsedSelf.project}/${parsedSelf.id}/${pathId}${
```
Dinika, line 489: **A note** on why I changed the `path`.

In my-data page in [staging](https://staging.nise.bbp.epfl.ch/nexus/) the download by selecting all rows in the table is not working because (according to the server response) the path is repeated for some resources that are being downloaded.

If we do a text search (ctrl + F) on the page, we see that there are 2 rows with column "name" = organization (this is one example). Apparently their project names are also same so the path was getting repeated. Adding the `id` obtained from `self` to the path makes the paths unique and fixes the problem.

I don't know if this should happen in production too. Let me know if you foresee issues with this solution.

src/shared/hooks/useAccessDataForTable.tsx
```diff
@@ -422,43 +432,94 @@ const fileNameForDistributionItem = (distItem: any, defaultName: string) => {
...
+    name: getResourceName(resource),
+    project: resource._project,
+    description: resource.description ?? '',
+    createdAt: resource._createdAt,
+    updatedAt: resource._updatedAt,
+    source: 'studios',
+    type: isArray(resource['@type'])
+      ? resource['@type']
+      : [resource['@type'] ?? ''],
+  };
```
bilalesi, line 455: why we take only the first one if the type is an array, 
I think the type is used to be showed in the data panel popup, so we can added all the types ?

src/shared/hooks/useAccessDataForTable.tsx
```diff
@@ -422,43 +432,94 @@ const fileNameForDistributionItem = (distItem: any, defaultName: string) => {
...
+      // Now store an object for each distribution item
+      resource.distribution.forEach((distItem, index) => {
+        localStorageObjs.push({
+          ...baseLocalStorageObject(resource, `${index}`),
+          distribution: {
+            hasDistribution: true, // So, we don't download the distribution twice
+            contentSize:
+              (distItem.contentSize as { value: number })?.value ?? // TODO: Verify
+              distItem.contentSize ??
+              0,
```
bilalesi, line 492: suggestion: we can use the get fn from lodash, it will give nice and typeless expression `_.get(distItem.contentSize, 'value', null) ?? distItem.contentSize`

src/shared/organisms/DataPanel/DataPanel.tsx
Dinika: Yeah, that was a good catch. I've fixed the path length issue in PR for [3898](https://github.com/BlueBrain/nexus-web/pull/1250).

src/shared/hooks/useAccessDataForTable.tsx
Dinika: Hmm, I'm taking all the elements in the array if `resource[@type]` is an array. All these types are also shown in the datapanel:
Screenshot from 2023-06-07 12-11-56

src/shared/hooks/useAccessDataForTable.tsx
Dinika: bon idee! :bow:

bilalesi (approved):

bilalesi (approved):

PR #1255, opened 2023-06-06
F 3907/delete project

bilalesi:
Fixes #https://github.com/BlueBrain/nexus/issues/3907

## Description

1. Change the design of the danger zone to show the deprecate / delete action sections.
2. Refactor the deprecate/delete modal.
3. Add the DELETE Api endpoint.

DangerZoneView
DeleteProjectModal
DeprecateProjectModal

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.


src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx
```diff
@@ -323,6 +341,7 @@ const OrganizationProjectsPage: React.FC<{}> = ({}) => {
           onCreateClick={() => updateCreateModelVisibility(true)}
           permissions={['projects/create']}
           path={[`/${orgLabel}`]}
+          supTitle="Organization"
```
Dinika, line 344: Can we rename `supTitle`. I may be wrong but it looks like the value of this is only ever `Organization`. If that's the case then maybe `Organization` or `OrgTitle` may be more descriptive (we could also make this a boolean instead and if it is true display the string "Organization" in the component). 

If there are other possible values for this, then I would suggest the name `RouteTitle`, `RouteLabel`, `AccessTitle` or something along those lines. The `sup` in `supTItle` is confusing for me.

src/shared/modals/DangerZone/DangerZoneAction.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  (projectName as string)?.toLowerCase() !== matchTerm;
+                return (
+                  <Button
+                    loading={status}
+                    disabled={disabled}
+                    type="primary"
+                    htmlType="submit"
+                    danger
+                  >
+                    I understand the consequences, {action} this project
```
Dinika, line 93: According to the types `action` can be undefined. In case it is, we probably don't want to display this text.

src/shared/modals/DangerZone/DangerZoneAction.tsx
```diff
@@ -0,0 +1,105 @@
+import React, { useState } from 'react';
+import { Button, Col, Form, Input, Modal, Row } from 'antd';
+
+export type TDangerZoneActionProps = {
+  open: boolean;
+  onClose(): void;
+  matchTerm: string;
+  handler(values: any): void;
+  action?: 'deprecate' | 'delete';
```
Dinika, line 9: Hmm, in which case can `action` be undefined? It seems weird to me that some component would call this component if there is no `action`.

Dinika (changes requested):

Dinika:
I was not able to compile the project locally. I get this error:
```
ERROR in /home/dinika/code/work/nexus-web/src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx
[tsl] ERROR in /home/dinika/code/work/nexus-web/src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx(289,9)
      TS2322: Type 'ProjectResponseCommon[]' is not assignable to type 'TProjectResponseCommonExtended[]'.
  Type 'ProjectResponseCommon' is not assignable to type 'TProjectResponseCommonExtended'.
    Property '_markedForDeletion' is missing in type 'ProjectResponseCommon' but required in type '{ _markedForDeletion: boolean; }'
```

src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx
```diff
@@ -48,6 +48,9 @@ const SHOULD_INCLUDE_DEPRECATED = true;
 type TOrganizationOptions = {
   orgLabel: string;
 };
+type TProjectResponseCommonExtended = ProjectResponseCommon & {
```
Dinika, line 51: ERROR in /home/dinika/code/work/nexus-web/src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx
[tsl] ERROR in /home/dinika/code/work/nexus-web/src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx(289,9)
      TS2322: Type 'ProjectResponseCommon[]' is not assignable to type 'TProjectResponseCommonExtended[]'.
  Type 'ProjectResponseCommon' is not assignable to type 'TProjectResponseCommonExtended'.
    Property '_markedForDeletion' is missing in type 'ProjectResponseCommon' but required in type '{ _markedForDeletion: boolean; }'

src/pages/OrganizationProjectsPage/OrganizationProjectsPage.tsx
bilalesi: the RouteHeader components is global component so it's preferably to not make special cases in it, 
and the idea behind the naming is like `<sup>` html tag, so it will be only on top of the title,
and since the component is `RouteHeader`, i think properties must not include the prefix of `Route`
since it's self explanatory

src/shared/modals/DangerZone/DangerZoneAction.tsx
bilalesi: for this if you see in the state that manage the actions, it must be empty (`undefined`) at the beginning if not the action will take either `deprecate` or `delete` and then action selecting the action will not be correct, 
the action property is not only for the text that it's showed but also for the action of the button, 
may be you are right if we have to separate the text from the `action`

Dinika (approved):

PR #1258, opened 2023-06-06
Hotfix/data download v1

bilalesi:
## Description
1. Use `ignoreNotFound` to the fetch `GET: /archive` bypass not founded resources.
2. Create a new type for distribution `TDistribution`.
3. Add `sizeCalculator` that take on consideration the different state of distribution and `contentSize`.
4. Add check of the distribution shape if an Array

## Types of changes

## Checklist:


Dinika (commented):
Some issues I found:

1. If I go to "my-data" section and select all, then download, the tar file that is downloaded looks corrupted to me. I only tested this on my linux laptop though, so if it is not an issue on your mac. Could you please double check this case.

2. When I go to "Neuron Electrophysiology" search page, and select more than one resources to download, the tar file only contains one file (even though each of those resources had some distributions).

PR #1260, opened 2023-06-08
3937 // Display name or label of resource in my-data if present other…

Dinika:
…wise id

When label, `preLabel`, or `name` of resource is present, display that, otherwise display id or self in my-data table.

Screenshot from 2023-06-08 18-21-26

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/molecules/MyDataTable/MyDataTable.tsx
+++ b/src/shared/molecules/MyDataTable/MyDataTable.tsx
@@ -21,6 +21,7 @@ import {
   removeLocalStorageRows,
   toLocalStorageResources,
 } from '../../../shared/utils/datapanel';
+import { getResourceLabel } from '../../../shared/utils';
 
 export const MAX_DATA_SELECTED_SIZE__IN_BYTES = 1_073_741_824;
 export const MAX_LOCAL_STORAGE_ALLOWED_SIZE = 4.5;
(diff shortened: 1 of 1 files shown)


src/shared/molecules/MyDataTable/MyDataTable.tsx
```diff
@@ -189,14 +190,15 @@ const MyDataTable: React.FC<TProps> = ({
...
             const { org, project } = makeOrgProjectTuple(record._project);
             return (
               <Tooltip title={text}>
                 <Button
                   style={{ padding: 0 }}
                   type="link"
-                  onClick={() => goToResource(org, project, text)}
+                  onClick={() => goToResource(org, project, resourceId)}
                 >
                   {showedText}
```
bilalesi, line 203: i think `showedText` will be replaced with `text` since we are getting the right `name` now
and the Tooltip will get the `@id` if there is no name

Dinika: Thanks for catching that. I've made the change. Here's how it looks like now:

Screenshot from 2023-06-12 10-05-34
When a name was found


Screenshot from 2023-06-12 10-05-55
When no name was found

bilalesi (approved):

PR #1279, opened 2023-06-19
fix: file extenstion from json files

bilalesi:
Fixes #

## Description
1- Get the correct extension of a file (first check the `filename` then the `encodingFormat`)
2- when filtering the `json` resource from the `json` files, the @type could be an array so filtering the by array must be added

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/organisms/DataPanel/DataPanel.tsx
+++ b/src/shared/organisms/DataPanel/DataPanel.tsx
@@ -466,7 +466,6 @@ const DataPanel: React.FC<Props> = ({}) => {
               ? resource.distribution?.encodingFormat[0]
               : resource.distribution?.encodingFormat
           );
-
           return {
             size,
             contentType: contentType?.toLowerCase(),
@@ -490,7 +489,6 @@ const DataPanel: React.FC<Props> = ({}) => {
 
     return groupBy(newDataSource, 'contentType');
   }, [dataSource]);
-
   const existedTypes = compact(Object.keys(resourcesGrouped)).filter(
     i => i !== 'undefined'
   );
@@ -511,7 +509,10 @@ const DataPanel: React.FC<Props> = ({}) => {
 
       if (key === 'json') {
         const metadataFiles = value.filter(
-          v => v?.localStorageType === 'resource' && v['@type'] !== 'File'
+          v =>
+            v?.localStorageType === 'resource' &&
+            v['@type'] !== 'File' &&
+            v['@type'].includes('File')
         ).length;
 
         // We don't want to display `json` for metadata files since they are always downloaded.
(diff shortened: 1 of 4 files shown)


Dinika (changes requested):

src/utils/contentTypes.ts
```diff
@@ -63,3 +63,18 @@ export const fileExtensionFromResourceEncoding = (encodingType?: string) => {
 
   return extension;
 };
+
+export const getCorrectFileExtension = (
```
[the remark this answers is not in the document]
bilalesi: by `correct` i mean: 
since we have different source to get the extension than we should get the correct one, because if we take from one source than it might be wrong (coz, we have another source for extracting that extension if it exist)

src/shared/organisms/DataPanel/DataPanel.tsx
```diff
@@ -460,8 +460,12 @@ const DataPanel: React.FC<Props> = ({}) => {
             Boolean(resource.distribution?.contentSize)
               ? 'File'
               : 'Resource';
-          const contentType =
-            resource.distribution?.label?.split('.')?.pop() ?? '';
+          const contentType = getCorrectFileExtension(
+            resource.distribution?.label ?? '',
```
Dinika, line 464: it cannot be the case that `resource.distribution.label` is undefined (because we tackle take care of that in `toLocalStorageResources` function). So we can remove `?? ''`

Dinika (approved):

src/shared/organisms/DataPanel/DataPanel.tsx
bilalesi: but it's possible that we can not have `distribution`  then resource.distribution will be `undefined`

Dinika: Well, actually, today it's not possible. The `toLocalStorageResources()` function normalizes the data such that there is always `distribution`

src/utils/contentTypes.ts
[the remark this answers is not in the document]
Dinika: In that case, I think you want to call the function `getNormalizedExtension()` since it takes care of considering the different sources and retrieving the extension from it.

If you have created a function or any utility you always want it to be "correct". The meaning of "correct" itself is determined by the implementation details of that function/utility and the tests you wrote, and hence  correctness is always implied.

I guess it'll be clearer if I give a counter example, would you ever create a function called "getWrongExtesion()" (except maybe in a test, when creating a mock)? Probably not.  Therefore the word "correct" is not useful, and potentially confusing in a name.

src/shared/organisms/DataPanel/DataPanel.tsx
bilalesi: make since, so the distribution object type must also change. 
I change the distribution type and the content type function

PR #1286, opened 2023-06-26
fix: add realm to identities to differntiate similair names and singu…

bilalesi:
…larity of reactKeys

Fixes #

## Description
In the ACLs & Permissions Tab, there might be multiple entities with the same name but they are in different realms,
We should enable the user to differentiate between user by their realms

Screenshot 2023-06-26 at 10 38 52

## Types of changes

## Checklist:

### DIFF
--- a/src/subapps/admin/components/Settings/PermissionsAclsSubView.tsx
+++ b/src/subapps/admin/components/Settings/PermissionsAclsSubView.tsx
@@ -1,6 +1,6 @@
 import React, { useMemo } from 'react';
 import { groupBy, sortBy } from 'lodash';
-import { Table, Collapse, Checkbox, Empty, Spin } from 'antd';
+import { Table, Collapse, Checkbox, Empty, Spin, Tag } from 'antd';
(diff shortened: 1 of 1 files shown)


src/subapps/admin/components/Settings/PermissionsAclsSubView.tsx
```diff
@@ -21,6 +21,7 @@ type DataType = {
   create?: boolean;
   query?: boolean;
   children?: DataType[];
+  realm?: string;
```
Dinika, line 24: I think it would make sense to create another TypeScript type for `children` that excludes `realm` to make this more cleaner. I'm thinking something along the lines of `type ChildType = Omit<DataType, 'realm'>`. We can ofcourse exclude more keys too if it makes sense.

This way we can make the code below clearer too by using typeguards to check if the row is for a parent or child.

Dinika (approved):
Left a comment. Also, I think we should add some unit tests.

PR #1287, opened 2023-06-26
fix: show full types in multiple lines in tooltip

bilalesi:
Fixes #

## Description
In the resources listing in my data page, the types of each resource might be an array, in the relative row, we show only the label of the type, in the tooltip we show the whole URI of the type so we should present them in a better way so each type will have a new line in the tooltip

Screenshot 2023-06-26 at 10 42 37

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/molecules/MyDataTable/MyDataTable.tsx
+++ b/src/shared/molecules/MyDataTable/MyDataTable.tsx
@@ -130,6 +130,24 @@ export const notifyTotalSizeExeeced = () => {
     key: 'data-panel-size-exceeded',
   });
 };
+const getTypesTrancated = (text: string | string[]) => {
+  let types = '';
+  let typesWithUrl = text;
+  if (isArray(text)) {
+    types = text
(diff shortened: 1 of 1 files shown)


src/shared/molecules/MyDataTable/MyDataTable.tsx
```diff
@@ -130,6 +130,24 @@ export const notifyTotalSizeExeeced = () => {
...
+const getTypesTrancated = (text: string | string[]) => {
+  let types = '';
+  let typesWithUrl = text;
+  if (isArray(text)) {
+    types = text
+      .map(item => (isValidUrl(item) ? item.split('/').pop() : item))
+      .join('\n');
+    typesWithUrl = text.join('\n');
+  } else if (isString(text) && isValidUrl(text)) {
+    types = text.split('/').pop() ?? '';
```
Dinika, line 142: The `?? ''` is not needed here, right? If there is no `/` in `text` we'll just get the whole string so the program will never reach `?? ''`.

PR #1289, opened 2023-06-26
hotfix: my data filters

bilalesi:
## Description
1. set createdAt as the default value of DateField
2. remove across project from the my-data page

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/molecules/MyDataTable/MyDataTable.tsx
+++ b/src/shared/molecules/MyDataTable/MyDataTable.tsx
@@ -87,6 +87,7 @@ type TProps = {
   sort: string[];
   updateSort(value: string[]): void;
   locate: boolean;
+  query: string;
 };
 export const makeOrgProjectTuple = (text: string) => {
   const parts = text.split('/');
@@ -171,6 +172,7 @@ const MyDataTable: React.FC<TProps> = ({
   sort,
   updateSort,
   locate,
+  query,
 }) => {
   const history = useHistory();
   const location = useLocation();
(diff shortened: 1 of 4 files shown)


src/shared/molecules/MyDataTable/MyDataTable.tsx
```diff
@@ -244,12 +246,15 @@ const MyDataTable: React.FC<TProps> = ({
           return (
             <div>
               organization / project
-              <Sorter
-                name="name"
-                order={orderDirection}
-                onSortAscend={() => updateSort(['_project'])}
-                onSortDescend={() => updateSort(['-_project'])}
-              />
+              {query ||
```
Dinika, line 249: Is this intentional? Wouldn't the remainder of this expression not evaluate if `query` is truthy? 

I think it's also hard to understand the intended logic here. Are we saying that if `query` is falsy only then show the sorter? If so, I think using the following pattern here might be clearer: `query ? (render something) : ( render something else)`

Dinika (changes requested):

src/shared/molecules/MyDataTable/MyDataTable.tsx
Dinika: I see. I still think that it might be easier to understand the following syntax more than the current one. Not a must have though, so I'll approve the PR.

{
      query.trim().length === 0 
           ? (<Sorter
                    name="name"
                    order={orderDirection}
                    onSortAscend={() => updateSort(['_project'])}
                    onSortDescend={() => updateSort(['-_project'])}
                  />)
           : query
}

Dinika (approved):

PR #1290, opened 2023-06-26
US 2 // Filter by projects

Dinika:
Allows users to filter resources in data explorer by projects
Screenshot from 2023-06-26 13-16-33

## Types of changes

## Checklist:

### DIFF
--- a/src/subapps/dataExplorer/ProjectSelector.tsx
+++ b/src/subapps/dataExplorer/ProjectSelector.tsx
@@ -1,0 +1,86 @@
+import { ProjectResponseCommon } from '@bbp/nexus-sdk';
+import { useNexusContext } from '@bbp/react-nexus';
+import { AutoComplete, Input, notification } from 'antd';
+import React, { useState } from 'react';
+import { useQuery } from 'react-query';
+import './styles.less';
+
+interface Props {
+  onSelect: (orgLabel?: string, projectLabel?: string) => void;
+}
+
+export const ProjectSelector: React.FC<Props> = ({ onSelect }: Props) => {
+  const { data: projects } = useProjects();
+  const [searchTerm, setSearchTerm] = useState('');
+
+  const allOptions = [
+    { value: AllProjects, label: AllProjects },
+    ...(projects?.map(projectToOption) ?? []),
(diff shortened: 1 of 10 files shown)


src/subapps/dataExplorer/ProjectSelector.tsx
```diff
@@ -0,0 +1,86 @@
...
+    queryKey: ['data-explorer-projects'],
+    retry: false,
+    queryFn: async () => {
+      // TODO: Replace this with aggregation API when it is ready.
+      // NOTE: At the moment it is not possible to get all projects a user has access to in 1 request. To get around this, we make 2 requests:
+      // 1st request -> Get the total number of projects (n) a user can see
+      // 2nd request -> Use the total retrieved from above request to specify the size of the projects to return.
+      const firstPageOfProjects = await nexus.Project.list(undefined, {
+        size: 1,
+      });
```
bilalesi, line 73: I think the total projects are not high, so i suggest to use predefined value  as 100, without using a handshake request just to get the total and as you said in the comment when the api will be ready we can have a clean way to do it

src/subapps/dataExplorer/ProjectSelector.tsx
```diff
@@ -0,0 +1,86 @@
...
+        className="search-input"
+        popupClassName="search-menu"
+      >
+        <Input.Search size="middle" />
+      </AutoComplete>
+    </div>
+  );
+};
+
+export const AllProjects = 'All projects';
```
bilalesi, line 50: can we rename it as constant (uppercase) so we can understand by the format (and it will be consistent as in different places in the APP)

src/subapps/dataExplorer/ProjectSelector.tsx
Dinika: If it's not a big deal, I would like to keep it as it is atm.

The number of projects in prod is 138 (so more than 100). The solution right now will work in all the cases, as it is, and will always show the *correct* result (at the cost of 1 extra api call that took 259 ms on VPN). Like you said, this will soon be replaced by aggregation api.

PR #1293, opened 2023-06-27
DE10: jump between nighbours resources

bilalesi:
Fixes #https://app.zenhub.com/workspaces/blue-brain-nexus-60ace035034cae00105c9307/issues/zh/10

## Description
1. Create a new component for navigation items
2. Create a new component for shrink-ed navigation 
3. Create Navigation stack
4. All the user to navigate between the resource by clicking on the navigation 
5. Save the navigation state to local-storage and populate it if the user refresh the page

## Types of changes

## Checklist:


src/shared/organisms/DataExplorerGraphFlowNavigationStack/NavigationStack.spec.tsx
```diff
@@ -0,0 +1,260 @@
...
+    user = userEvent.setup();
+  });
+  it('should render the correct number of NavigationStackItem components in the state', () => {
+    const navigationItems = container.querySelectorAll(
+      '.navigation-stack-item:not(.no-more)'
+    );
+    expect(navigationItems.length).toBe(2);
+  });
+  it('should render the correct number of NavigationStackItem components after hit the return back btn', () => {
+    store.dispatch(ReturnBackDataExplorerGraphFlow());
```
Dinika, line 125: It would be better to get the button in the `NavigationBack` component here and click that instead of manually dispatching an action. This way we test user bahavior (i.e. clicking the button resets the stack) instead of implementation detail (dispatching this particular action resets the stack).

src/shared/organisms/DataExplorerGraphFlowNavigationStack/NavigationStack.spec.tsx
```diff
@@ -0,0 +1,260 @@
...
+    rerender(app);
+    const navigationItemsAfterBack = container.querySelectorAll(
+      '.navigation-stack-item:not(.no-more)'
+    );
+    expect(navigationItemsAfterBack.length).toBe(1);
+  });
+  it('should render the correct number of NavigationStackItem after multiple navigation', () => {
+    store.dispatch(
+      AddNewNodeDataExplorerGraphFlow({
```
Dinika, line 134: It might be a good idea to create a utility function in the test that gives back a navigation item. The test will read cleaner that way:

```javascript
const addItemToNavigationHistory = (resource) => {
    const navigationItem = {
         isDownloadable: false,
        _self:
          'https://bbp.epfl.ch/nexus/v1/resources/${resource[0]}/${resource[1]}/_/${resource[2]}',
        title: resource[2],
        types: [],
    }
}
store.dispatch(AddNewNodeDataExplorerGraphFlow(resource))

// it block
addItemToNavigationHistory( ['bbp', 'atlas', 'https://bbp,neuroshaes.org']);
addItemToNavigationHistory( ['bbp', 'atlas', 'https://bbp,neuroshaes.org']);

```

src/shared/organisms/DataExplorerGraphFlowNavigationStack/NavigationStack.spec.tsx
```diff
@@ -0,0 +1,260 @@
...
+    const navigationHiddenItems = container.querySelectorAll(
+      '.navigation-stack-item:not(.no-more):not(.more)[hidden]'
+    );
+    expect(navigationHiddenItems.length).toBe(4);
+  });
+  it('should show the collapse button when the NavigationStackShrinkedItem is clicked', () => {
+    const navigationStackShrinkedItem = container.querySelector(
+      '.navigation-stack-item.more'
+    );
+    if (navigationStackShrinkedItem) {
```
Dinika, line 216: This looks wrong. Why would `navigationStackShrinkedItem` be falsy? If it is, this test is not really doing anything. We should set this test up in a way that `navigationStackShrinkedItem` is always in the DOM and then continue with the assertions.

src/shared/organisms/DataExplorerGraphFlowNavigationStack/NavigationStack.spec.tsx
```diff
@@ -0,0 +1,260 @@
...
+      expect(collapseBtn).not.toBeNull();
+      expect(collapseBtn).toBeInTheDocument();
+      expect(store.getState().dataExplorer.shrinked).toBe(false);
+    }
+  });
+  it('should hide the collapse button when the collapse button is clicked', () => {
+    const navigationCollapseButton = container.querySelector(
+      '.navigation-collapse-btn'
+    );
+    if (navigationCollapseButton) {
```
Dinika, line 228: same here

src/shared/organisms/DataExplorerGraphFlowNavigationStack/NavigationStack.spec.tsx
```diff
@@ -0,0 +1,260 @@
...
+      user.click(navigationCollapseButton);
+      const collapseBtn = container.querySelector('.navigation-collapse-btn');
+      expect(collapseBtn).toBeNull();
+      expect(store.getState().dataExplorer.shrinked).toBe(true);
+    }
+  });
+  it('should the items in the stack be 4 when the user jump to the 5th item', () => {
+    store.dispatch(JumpToNodeDataExplorerGraphFlow(4));
```
Dinika, line 236: I'm a bit confused. How is it possible to jump to node 4, if the initialState only has 2 nodes? We should clean the `state` after each test so that the tests don't share any preserved state.

Dinika: If I comment out the test `should render the NavigationStackShrinkedItem when it passed MAX_NAVIGATION_ITEMS_IN_STACK` then the test below for `should the fourth item in the stack will be the current one when the user jump to the 4th item` starts failing. This signals that the tests are right now dependant on the sequence. This can lead to flakiness. We should fix this (by resetting state in beforeEach or cleaning it in afterEach) and make tests more predictable.

Dinika (changes requested):

Dinika:
Testing wise, things look good.

src/shared/canvas/DataExplorerGraphFlow/DateExplorerGraphFlow.tsx
```diff
@@ -1,10 +1,68 @@
-import React from 'react';
+import React, { useRef, useEffect, CSSProperties } from 'react';
```
Dinika, line 1: Something I was a bit confused about is that this PR includes changes for user stories 10 and 11. I thought each user story would be the most atomic unit of work that we will do, therefore 1 PR per user story. Maybe this is something we should discuss when we decide on how the user stories are going for us.

src/shared/organisms/DataExplorerGraphFlowNavigationStack/NavigationStack.spec.tsx
bilalesi: why it is wrong, we can test the `navigationStackShrinkedItem` if it's in the document but it will still a possibility to be null, due the response from `querySelector ` and for the test to complete we should have it in the document to continue

src/shared/canvas/DataExplorerGraphFlow/DateExplorerGraphFlow.tsx
bilalesi: It's okey to work on 2 user story in one PR, and it was already mentioned by Deneil, I can two PRs for this since the components are pretty much the same We can make it in one PR.

Dinika (approved):

PR #1294, opened 2023-06-27
fix: when there is no query don't render the sorter

bilalesi:
Fixes #

## Description
When the user start typing a query in the search box; the sorter must be hidden due the delta API  do not support the query and the sorter to be in one query

## Types of changes

## Checklist:


Dinika (approved):
I really think we should unit test this.

PR #1297, opened 2023-06-27
Hotfix: allow anonymous users to access all routes

bilalesi:
## Types of changes

## Checklist:

### DIFF
--- a/src/shared/views/SubAppsView.tsx
+++ b/src/shared/views/SubAppsView.tsx
@@ -1,7 +1,6 @@
 import { Location } from 'history';
 import * as React from 'react';
 import { Redirect, Route, Switch, useLocation } from 'react-router-dom';
-import PrivateRoute from '../../shared/organisms/ProtectedRoute/ProtectedRoute';
 import NotFound from './404';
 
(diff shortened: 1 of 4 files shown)


Dinika (changes requested):
There is some redirection logic in `Login.tsx` and `auth.ts` files. Should we remove those?

src/shared/views/SubAppsView.tsx
```diff
@@ -1,7 +1,6 @@
 import { Location } from 'history';
```
Dinika, line 1: I can't remember where we decided to leave comments from tests, so I'm leaving them here.

When the user is not logged in and they visit http://localhost:8000/search?layout=Neuron%20Electrophysiology or http://localhost:8000/my-data the screen looks a bit weird because the top of it is not fully visible:
Screenshot from 2023-06-27 13-28-49

Screenshot from 2023-06-27 13-27-50

Dinika (changes requested):

bilalesi:
good spot, i will fix the style

Dinika (approved):

Dinika (approved):

PR #1302, opened 2023-06-30
US 6 // Show resources that contain a specific value

Dinika:
Fixes #[User Story 6](https://app.zenhub.com/workspaces/blue-brain-nexus-60ace035034cae00105c9307/issues/zh/6)

## Description

Allows user to filter rows that contain a value provided by them
Screenshot from 2023-07-03 15-00-47

## Types of changes

## Checklist:


src/shared/layouts/FusionMainLayout.less
```diff
@@ -5,13 +5,18 @@
   height: 100%;
   min-height: 100vh !important;
 
+  &.wall {
+    .fusion-main-layout__content {
+      margin-top: 0;
+    }
+  }
```
bilalesi, line 12: this one we can remove it so it does not appear again in merge conflicts
also it's better to remove the `classname` that it depends on it in the `MainLayout`

bilalesi (approved):

PR #1307, opened 2023-07-04
DE12: Enable File download

bilalesi:
## Description
1. Enable download file when resolution detect a File
2. Preserve the page when tabs are navigated inside the resource view
3. Fix the margin at the top (since design has changed)

Dowload File from Editor

## Types of changes

## Checklist:


src/__mocks__/handlers/DataExplorer/handlers.ts
```diff
@@ -0,0 +1,761 @@
...
+  _rev: 2,
+  _schemaProject:
+    'https://bbp.epfl.ch/nexus/v1/projects/neurosciencegraph/datamodels',
+  _self:
+    'https://bbp.epfl.ch/nexus/v1/resources/public/sscx/datashapes:neuronmorphology/neuronmorphologies%2Fbfdd4d1a-8b06-46fe-b663-7d9f8020dcaf',
+  _updatedAt: '2023-06-23T07:34:56.011Z',
+  _updatedBy: 'https://bbp.epfl.ch/nexus/v1/realms/bbp/users/cgonzale',
+};
+
+const initialResource = {
```
Dinika: Because mock data can be fairly large, it's often a good idea to create utility functions that return them. This way, repetitions of large mock values are removed (there's only one function that contains the whole data). 

If you want, there's already some utility functions (called `getMockResource()`) that can be resued here maybe.

I haven't yet verified if we exclude handler files from the production build, but if we are not excluding them, then we should be specially careful when adding large mock objects.

src/shared/components/ResourceEditor/editorUtils.ts
```diff
@@ -76,22 +99,32 @@ export async function resolveLinkInEditor({
...
             _self: data._self,
             title: getResourceLabel(data),
             types: getNormalizedTypes(data['@type']),
-            resource: [
-              entity?.orgLabel,
-              entity?.projectLabel,
-              data['@id'],
-              data._rev,
-            ],
+            resource: isDownloadable
```
Dinika, line 114: I like the naming `isDownloadable`. It made this file change very easy to understand in the context of the PR :star:

src/shared/components/ResourceEditor/editorUtils.ts
```diff
@@ -104,47 +137,47 @@ export async function resolveLinkInEditor({
         data = await nexus.Resource.list(undefined, undefined, {
           locate: url,
         });
+        if (
+          !data._total ||
+          (!data._total && url.startsWith('https://bbp.epfl.ch'))
```
Dinika, line 142: I think the repetition of `!data.total` before and after `||` is not needed (because if it was false, then it will be false after || check also). Maybe you want to check instead `!data.total || !url.startsWith('https://bbp.epfl.ch')`

As a separate issue, since we are only checking for the production domain, we will not be able to test this in other environments like dev, staging etc.

src/shared/containers/ResourceEditor.tsx
```diff
@@ -119,23 +123,27 @@ const ResourceEditorContainer: React.FunctionComponent<{
...
+      dispatch(InitDataExplorerGraphFlowLimitedVersion(true));
+    } else {
+      dispatch(
+        InitNewVisitDataExplorerGraphView({
+          current: {
+            _self: data._self,
+            types: getNormalizedTypes(data['@type']),
+            title: getResourceLabel(data),
+            resource: [
+              orgProject?.orgLabel ?? '',
```
Dinika, line 136: It would be nice to have a utility that returns this resource array. I can foresee cases where we accidentally changes the sequence of elements (for example, project before organization).

Dinika (approved):
Looks good. The comments are "nice to have"s not "must haves"

src/shared/components/ResourceEditor/editorUtils.ts
bilalesi: yeah the `isExternalLink` will be extended to test dev/staging links @Dinika

PR #1308, opened 2023-07-04
US 5 // Filter resources that have path

Dinika:
Screenshot from 2023-07-04 14-24-42

## Types of changes

## Checklist:

### DIFF
--- a/src/subapps/dataExplorer/DataExplorer.tsx
+++ b/src/subapps/dataExplorer/DataExplorer.tsx
@@ -4,41 +4,23 @@ import { notification } from 'antd';
 import { isString } from 'lodash';
 import React, { useReducer } from 'react';
 import { useQuery } from 'react-query';
-import { getResourceLabel } from '../../shared/utils';
 import { DataExplorerTable } from './DataExplorerTable';
 import './styles.less';
 import { ProjectSelector } from './ProjectSelector';
-import {
-  CONTAINS,
-  EMPTY_VALUE,
-  PredicateFilterT,
-  PredicateSelector,
-  doesResourceContain,
-  isPathMissing,
-} from './PredicateSelector';
-import { normalizeString } from 'utils/stringUtils';
+import { PredicateSelector } from './PredicateSelector';
 
 export interface DataExplorerConfiguration {
   pageSize: number;
   offset: number;
   orgAndProject?: [string, string];
-  predicatePath: string | null;
-  predicateFilter: PredicateFilterT | null;
-  predicateValue: string | null;
+  predicateFilter: ((resource: Resource) => boolean) | null;
 }
 
 export const DataExplorer: React.FC<{}> = () => {
   const nexus = useNexusContext();
 
   const [
-    {
-      pageSize,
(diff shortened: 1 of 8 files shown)


src/subapps/dataExplorer/PredicateSelector.tsx
```diff
@@ -142,11 +176,47 @@ const getPathsForResource = (
...
+
+  return criteria === 'exists' ? false : true;
+};
+
 export const isPathMissing = (
   resource: { [key: string]: any },
   path: string
 ): boolean => {
-  if (path in resource) {
+  if (path in resource && path !== '') {
```
bilalesi, line 219: Is there situation where the path is empty string ?
If so is it better to filter paths when you generate them (just a suggestion)

src/subapps/dataExplorer/PredicateSelector.tsx
```diff
@@ -91,9 +119,15 @@ export const PredicateSelector: React.FC<Props> = ({
...
+export const DOES_NOT_EXIST = 'Does not exist';
+export const EXISTS = 'Exists';
 export const CONTAINS = 'Contains';
-export type PredicateFilterT = typeof EMPTY_VALUE | typeof CONTAINS | null;
+
+export type PredicateFilterT =
+  | typeof DOES_NOT_EXIST
+  | typeof EXISTS
+  | typeof CONTAINS
+  | null;
```
bilalesi, line 130: I suggest use this to be more `Typescripty`
`const PREDICATE_VALUES = [DEFAULT_OPTION,  DOES_NOT_EXIST, EXISTS, CONTAINS] as const`;
`type TPredicateValues = typeof PREDICATE_VALUES;`
and then 
`type TPredicateValue = TPredicateValues[0 | 1 | 2] or TPredicateValues[number]`
and then select by index when you need it `TPredicateValues[index]`

src/subapps/dataExplorer/DataExplorer.tsx
```diff
@@ -4,41 +4,23 @@ import { notification } from 'antd';
...
+import { PredicateSelector } from './PredicateSelector';
 
 export interface DataExplorerConfiguration {
   pageSize: number;
   offset: number;
   orgAndProject?: [string, string];
-  predicatePath: string | null;
-  predicateFilter: PredicateFilterT | null;
-  predicateValue: string | null;
+  predicateFilter: ((resource: Resource) => boolean) | null;
```
bilalesi, line 16: I think it make sense if the `predicateFilter` can be rename it to `predicate`

bilalesi (approved):

src/subapps/dataExplorer/PredicateSelector.tsx
Dinika: bon idee! I'll do that in a new PR.

src/subapps/dataExplorer/DataExplorer.tsx
Dinika: this too.

PR #1309, opened 2023-07-04
US 7 // Filter resource that dont contain value

Dinika:
In case of arrays, if any one element does not contain the value, then the resource row is said is displayed.

Screenshot from 2023-07-04 16-45-47

## Types of changes

## Checklist:


src/shared/components/ResourceEditor/CodeEditor.tsx
```diff
@@ -35,7 +35,7 @@ const CodeEditor = forwardRef<codemiror.Editor | undefined, TCodeEditor>(
     return (
       <Spin spinning={busy}>
         <CodeMirror
-          data-testId="code-mirror-editor"
+          data-testid="code-mirror-editor"
```
bilalesi, line 38: yeah good one, react only accept custom attribute in lowercase

bilalesi (approved):

PR #1313, opened 2023-07-05
Fix: add referer to data explorer to allow return back to origin DEGF

bilalesi:
## Description
1. Allow the user to return back to the origin of the data explorer graph flow experience
2. Add the `referer` to data explorer state

## Types of changes

## Checklist:


Dinika (approved):
Would be nice to have a test for this case because I completely didn't think about this.

bilalesi:
> Would be nice to have a test for this case because I completely didn't think about this.

absolutely will be in another PR

PR #1317, opened 2023-07-07
US 4022 / Show count in a clearer format

Dinika:
Fixes #US 4022

Shows large numbers in the count of data explorer, as a comma separated value. For example the number 911899 will be shown as "911,899". Also makes the wording of count clearer.

Screenshot from 2023-07-11 13-15-35

## Types of changes

## Checklist:

### DIFF
--- a/src/subapps/dataExplorer/DatasetCount.tsx
+++ b/src/subapps/dataExplorer/DatasetCount.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import * as pluralize from 'pluralize';
+import { isNil } from 'lodash';
+import './styles.less';
+
+interface Props {
+  nexusTotal: number;
+  totalFiltered?: number;
+  totalOnPage: number;
+}
+
+export const DatasetCount: React.FC<Props> = ({
+  nexusTotal,
+  totalOnPage,
+  totalFiltered,
+}: Props) => {
+  return (
+    <div className="data-explorer-count">
+      <span>
+        Total:{' '}
+        <b>
+          {nexusTotal.toLocaleString(`en-US`)}{' '}
+          {pluralize('dataset', nexusTotal ?? 0)}
+        </b>{' '}
+      </span>
+
+      <span>
+        Sample loaded for review: <b>{totalOnPage}</b>
+      </span>
+
+      {!isNil(totalFiltered) && (
+        <span>
+          Filtered: <b>{totalFiltered}</b>
+        </span>
+      )}
+    </div>
+  );
+};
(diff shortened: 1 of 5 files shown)


src/subapps/dataExplorer/DatasetCount.tsx
```diff
@@ -0,0 +1,37 @@
...
+  nexusTotal,
+  totalOnPage,
+  totalFiltered,
+}: Props) => {
+  return (
+    <div className="data-explorer-count">
+      <span>
+        Total:{' '}
+        <b>
+          {nexusTotal.toLocaleString()} {pluralize('dataset', nexusTotal ?? 0)}
```
bilalesi, line 22: I think the clearer format is to show the large number with separator so the user can read the number
you can use built-in js fn 
```
new Intl.NumberFormat(locale, { 
      maximumSignificantDigits: 3,
}).format(Number(nexusTotal))}
where locale is 'de-CH' or any other locale
```

bilalesi (approved):

src/subapps/dataExplorer/DatasetCount.tsx
Dinika: The thing is `Number.prototype.toLocaleString` calls Intl.NumberFormat internall ([MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)) and it has a slimmer api so I went with that. I don't mind changing it to `Intl.NumberFormat` though

I think I should not pass the `maximumSignificantDigits` option because it round off the numbers:

```javascript
new Intl.NumberFormat('en-US', { 
      maximumSignificantDigits: 3,
}).format(Number(1234))
>> 1,230
```

The CH locales all look funny to me. For example, for the number 1343234325:

```javascript
new Intl.NumberFormat('fr-CH').format(Number(1343234325)) # "1 343 234 325"
new Intl.NumberFormat('de-CH').format(Number(1343234325)) # "1’343’234’325" 
```

Therefore, I am planning to use `en-US`:
```javascript
new Intl.NumberFormat('en-US').format(Number(1343234325)) # "1,343,234,325" 
```

bilalesi: yeah `toLocaleString` call the` Intl.NumberFormat` but does not change the format, do I miss smthg ?
and for the rounding is not a problem for me, most statistics value are showed with rounding
Screenshot 2023-07-11 at 13 39 37
for the locale, yeah looks funny, we just need to choose one, and stick with it across the app, i use `de-CH` in my data page, but you can change it as you prefer.

Dinika: 1. O yeah, a locale should be passed to `toLocaleString()`:
```
(121243).toLocaleString("de-CH")  # "121’243"
(121243).toLocaleString("en-US")  # "121,243"
```

2. I really think we should not round off the numbers. (If there are 1234 results, I shouldn't show there were only 1230 results returned from nexus), but that's just my opinion. If you want I can double check with Sam/Loris.

bilalesi: ahh okey, so we can use it without rounding

PR #1321, opened 2023-07-11
De14: Show popover info on hover

bilalesi:
Fixes # [US 14](https://app.zenhub.com/workspaces/blue-brain-nexus-60ace035034cae00105c9307/issues/zh/14)

## Description
1. Add hover to links to show the resolved data.
2. Add LRU cache to reduce requests to delta when the user keep hovering.
3. Add popover when click  on the link that resolved to multiple resources.
4. Use event listener and pure HTML to build the Popover and Tooltip.

https://github.com/BlueBrain/nexus-web/assets/2632473/7048e239-0fcf-4de9-bbea-40c8017e13c9
## Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


src/shared/components/ResourceEditor/index.tsx
```diff
@@ -199,6 +175,26 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
...
+  });
+  useEditorPopover({
+    orgLabel,
+    projectLabel,
+    ref: codeMirorRef,
+  });
+
+  React.useEffect(() => {
+    return () => {
+      if (location.pathname !== DATA_EXPLORER_GRAPH_FLOW_PATH) {
```
Dinika, line 192: Hmm, I like the idea of clearing the cache, but I'm not sure that we should clear the cache everythime the user navigates away from `data-explorer/graph-flow`. I think we should clear the cache only if the user navigates away from `data-explorer` (i.e. `!location.pathname.includes('data-explorer')`). Wdyt?


*Reason:*


Based on the user tests, I have a feeling that the user will often click on row in data-explorer, check out the graph-flow, come back to data-explorer table, then click another or same row again. There's a high chance that a lot of same links would need to be resolved but since the cache is cleared, we won't benefit from it.

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
...
+} from './editorUtils';
+import { TDELink } from '../../store/reducers/data-explorer';
+import { RootState } from '../../store/reducers';
+import useResolutionActions from './useResolutionActions';
+
+const downloadImg = require('../../images/DownloadingLoop.svg');
+
+export const CODEMIRROR_HOVER_CLASS = 'CodeMirror-hover-tooltip';
+export const CODEMIRROR_LINK_CLASS = 'fusion-resource-link';
+type TTooltipCreator = Pick<
```
Dinika, line 20: Isn't type `TTooltipCreator` the same as `TReturnedResolvedData`? Maybe we can reuse it instead of creating a new one?

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
...
+  nodeTag.className = 'tag';
+  tag && nodeTag.appendChild(document.createTextNode(tag));
+  tooltipItemContent.appendChild(nodeTag);
+
+  const nodeTitle = document.createElement('span');
+  nodeTitle.className = 'title';
+  nodeTitle.appendChild(document.createTextNode(title));
+  tooltipItemContent.appendChild(nodeTitle);
+
+  const nodeDownload = isDownloadable && document.createElement('img');
```
Dinika, line 60: Lines 60 to 63 can be wrapped inside a `if (isDownloadable)` block to avoid the repetition of `nodeDownload &&`

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
...
+    function removeTooltipFromDom(tooltip: HTMLDivElement) {
+      if (tooltip.parentNode) {
+        tooltip.parentNode.removeChild(tooltip);
+      }
+    }
+    function hideTooltip(tooltip: HTMLDivElement) {
+      if (!tooltip.parentNode) {
+        return;
+      }
+      setTimeout(() => {
```
Dinika, line 187: Shouldn't we clear this setTimeout too? Maybe this function can return the id it's the caller's (in this case the function `hide`) responsibility to clear it.

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
...
+        tooltip.style.top = `${editorRect.bottom}px`;
+      }
+      tooltip.style.left = `${editorRect.left}px`;
+    }
+    function removeTooltipFromDom(tooltip: HTMLDivElement) {
+      if (tooltip.parentNode) {
+        tooltip.parentNode.removeChild(tooltip);
+      }
+    }
+    function hideTooltip(tooltip: HTMLDivElement) {
```
Dinika, line 183: I find the abstractions in this useEffect somewhat confusing. We have 3 functions:
i) `hideTooltip`
ii) `removeTooltip`
iii) `hide`

They all sound like they should be hiding a tooltip so it's confusing why we need 3 of these. Most certainly we can combine `hideTooltip` and `removeTooltipFromDom` into one function (otherwise the abstraction seems too shallow, I'm afraid). 
The last function `hide` maybe does more than just hide the tooltip, so maybe a better name could be `cleanup`?

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
...
+        node.removeEventListener('mouseout', hide);
+        node.removeEventListener('click', hide);
+        node.removeEventListener('scroll', hide);
+      }
+
+      node.addEventListener('mouseout', hide);
+      node.addEventListener('click', hide);
+      node.addEventListener('scroll', hide);
+
+      const pool: ReturnType<typeof setTimeout> = setTimeout(() => {
```
Dinika, line 212: 1. Instead of `ReturnType<typeof setTimeout>` we could just say `number` because I think it's a relatively well known fact that setTimeout returns a number. (I'd even say that a type is not strictly necessary here because it's pretty clear it's always going to be a number, that might be too opinionated though).

2. Why `pool`? (As in, it's a pool of what? It looks like it's just a single timeout, not a pool of many timeouts). What about something like `timeoutId` instead? 

3. I'm not clear about why we need this setTimeout. The `hide` function already takes care of removing the event listeners, and calls the `hideTooltip` too.

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
...
+            }
+          });
+        }
+      }
+    }
+    // allow the tooltip only when the editor is not in edition mode
+    // and the popover is not open
+    allowTooltip && editorWrapper.addEventListener('mouseover', onMouseOver);
+    // remove the event listener when not allwoed
+    !allowTooltip &&
```
Dinika, line 267: Hmm why this? In the previous line aren't we only adding the listener when the allowTip is true? If so, there's no need to remove it, since it was never added.

src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -0,0 +1,405 @@
+import * as React from 'react';
```
Dinika, line 1: I think the functions inside this file, especially the two hooks, should be unit tested against different scenarios.

Dinika (approved):
Looks good. I've added small comments (mostly suggestions). I really think we should add some tests too for this. 
I'm starting manual testing now.

src/shared/components/ResourceEditor/ResourcesLRUCache.ts
```diff
@@ -0,0 +1,85 @@
+// NOTE: This file will be removed when delta introduce http cache headers
+import { NexusClient, Resource, PaginatedList } from '@bbp/nexus-sdk';
+import LRUCache from 'lru-cache';
+
+// TODO: Use nexus.httpGet to prepare for using http cache headers
```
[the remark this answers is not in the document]
bilalesi: currently I am using the LRU cache to cache the requests on the front due the multiple hover  can be done in the same link, I already request delta improvement to enable the client to cache the request in the browser using http headers

src/shared/components/ResourceEditor/useEditorTooltip.tsx
bilalesi: yeah, it was just a mess up trying to refactor removing from DOM also for popover,

src/shared/components/ResourceEditor/useEditorTooltip.tsx
bilalesi: 1. The returned type of the `setTimeout` is a number but actually is a `Timer` that has many properties,
 the real type is `NodeJS.Timeout` that implement a `Timer`, I was able to make the type  `NodeJS.Timeout` but I don't confuse the developer by a NodeJS type in React. (make it `number` does't hurt also').
 
2. it's Okey to name it a pool since `setTimeout` is using a pool of ids but i will rename it for clarity

3. A tooltip is meant to be short-live component so if the browser does not capture any of the events listed in the hide  (cleanup) fn that we should destroy the DOM node

src/shared/components/ResourceEditor/useEditorTooltip.tsx
bilalesi: This one is to remove it, there is two cases where we have to remove the listener
1. The component is un-mounting  
2. The editor is in the editing mode, then hover event must be removed

so this one is the one that remove the event listener when the user start editing
and the one in the cleanup function for the un-mounting

src/shared/components/ResourceEditor/useEditorTooltip.tsx
bilalesi: agreed, will be in different PR if you don't mind

PR #1323, opened 2023-07-13
US 4038 // Show only user data by default

Dinika:
Fixes #4038

1. Adds a toggle button that can be used to show or hide nexus metadata columns
2. Sorts columns and paths such that user columns appear before metadata columns
3. Gets expanded resource for data explorer

Screenshot from 2023-07-13 11-57-54

### DIFF
--- a/src/subapps/dataExplorer/styles.less
+++ b/src/subapps/dataExplorer/styles.less
@@ -12,6 +12,13 @@
     margin-bottom: 28px;
   }
 
+  .flex-container {
+    display: flex;
+    max-width: 90vw;
+    justify-content: space-between;
+    align-items: center;
+  }
+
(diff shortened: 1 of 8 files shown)


src/subapps/dataExplorer/styles.less
```diff
@@ -12,6 +12,12 @@
     margin-bottom: 28px;
   }
 
+  .flex-container {
+    display: flex;
+    max-width: 90vw;
+    justify-content: space-between;
+  }
```
bilalesi, line 19: can we have also `align-items: center` just so both sides of flex container not showed shifted since the toggle is bigger than the text in the left

bilalesi (approved):

PR #1331, opened 2023-07-17
F4048: Acl check actions

bilalesi:
Fixes #https://github.com/BlueBrain/nexus/issues/4048

## Description
In the JSON editor in resource view page, the save button was dependent on the user login state to show the button, 
the right way is the wrap the button with the permission config so the user will have the right to update the resource depends on his/her permissions.

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/components/ResourceEditor/index.tsx
+++ b/src/shared/components/ResourceEditor/index.tsx
@@ -7,7 +7,7 @@ import {
   SaveOutlined,
 } from '@ant-design/icons';
 import { useSelector } from 'react-redux';
-import { useNexusContext } from '@bbp/react-nexus';
+import { AccessControl } from '@bbp/react-nexus';
 import codemiror from 'codemirror';
 
 import 'codemirror/mode/javascript/javascript';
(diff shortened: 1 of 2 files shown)


src/shared/components/ResourceEditor/index.tsx
```diff
@@ -261,7 +261,11 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
                   style={switchMarginRight}
                 />
               )}
-              {userAuthenticated && (
+              <AccessControl
+                path={[`${orgLabel}/${projectLabel}`]}
+                permissions={['resources/write']}
```
Dinika, line 266: Would be nice to have these permissions as an enum and more typesafety somewhere to avoid chances of typos because right now if I pass `['abc']` to permissions, I don't get any errors.

Dinika (approved):

src/shared/components/ResourceEditor/index.tsx
bilalesi: the component is pre-built in the SDK, will do that when we got the SDK part (🤷🏻)

PR #1332, opened 2023-07-17
Hotfix: Fix back arrow location

bilalesi:
Fixes #
It's an autocomplete issue, when auto import the library, location was not imported from `useLocation`
so in this PR, i fixed the location object import
## Types of changes

## Checklist:


src/shared/organisms/DataExplorerGraphFlowNavigationStack/useNavigationStack.ts
```diff
@@ -13,6 +13,7 @@ import {
 const useNavigationStackManager = () => {
   const dispatch = useDispatch();
   const history = useHistory();
+  const location = useLocation();
```
Dinika, line 16: o wow! I'm surprised we didn't catch it before.

Dinika (approved):

PR #1333, opened 2023-07-17
Hotfix: Anchor icon path

bilalesi:
Fixes #

## Description
the anchor link icon in the JSON editor seems not showed correctly due the import url

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/components/ResourceEditor/ResourceEditor.less
+++ b/src/shared/components/ResourceEditor/ResourceEditor.less
@@ -78,13 +78,10 @@
       cursor: not-allowed !important;
     }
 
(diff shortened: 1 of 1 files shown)


Dinika:
seems to be working for me locally. Don't know how we missed this :disappointed: Since the icon is in css, testing would not catch it. We could either do a [screenshot test](https://docs.cypress.io/api/commands/screenshot) to catch this in future or add the icon in the javascript (the icon could be a simple, presentational icon, although I think it might make sense to make it clickable too) and add a simple test to check they exist.

Dinika (approved):

PR #1336, opened 2023-07-18
F3945: Filter by type mydata page

bilalesi:
Fixes #https://github.com/BlueBrain/nexus/issues/3945

## Description
1. User can do filter by type
2. User can search in the type filter
3. User able to select multiple types in the same time (blocked by delta, ui implmented)

https://github.com/BlueBrain/nexus-web/assets/2632473/3556c5a8-b4a8-4084-91f2-3c98bdf69d06

## Types of changes

## Checklist:


src/shared/canvas/MyData/MyData.tsx
```diff
@@ -116,6 +76,7 @@ const HomeMyData: React.FC<{}> = () => {
       ? `${dateField}-${dateFilterType}-${dateFilterRange}`
       : undefined;
   const order = sort.join('-');
+  const resourceTypes = types?.map(item => get(item, 'value'));
```
Dinika, line 79: I think `types` is always defined now. Atleast I don't see any place where it is being set to to a falsy value. So, prehaps we can remove the `?` check here and update the TS types for it.

src/shared/canvas/MyData/types.ts
```diff
@@ -7,7 +7,7 @@ export type TFilterOptions = {
   query: string;
   dateFilterType?: TDateFilterType;
   dateField: TDateField;
-  dataType?: string[];
+  types?: TType[];
```
Dinika, line 10: If we include the first change I suggest, we can remove the `?` from type definiton here too.

src/shared/canvas/MyData/types.ts
```diff
@@ -45,3 +45,26 @@ export type TDate = {
...
+export type TTypeAggregationsResult = {
+  '@context': string;
+  total: number;
+  aggregations: {
+    projects: TTypesAggregatedProperty;
+    types: TTypesAggregatedProperty;
+  };
+};
+
+export type TTypesAggregatedProperty = {
```
Dinika, line 64: Nice to have: There are `AggregatedProperty` and `AggregationsResult` types defined in `DataExplorerUtils.tsx`  file. They look very similar to the `TTypesAggregatedProperty` and `TTypeAggregationsResult`. We can remove one of them and use the same types in both places.

src/shared/molecules/MyDataHeader/MyDataHeader.tsx
```diff
@@ -1,433 +1,22 @@
...
-          style={{ marginLeft: 'auto' }}
-        />
-        <div className="filter-options">
-          <Checkbox checked={locate} onChange={onSearchLocateChange}>
-            <span className="locate-text">By resource id or self</span>
-          </Checkbox>
-        </div>
-      </div>
+      <span>{text}</span>
+      <span>{total ? `${prettifyNumber(total)} ${label}` : ''}</span>
```
Dinika, line 12: nice

src/shared/molecules/MyDataHeader/MyDataHeaderFilters/TypeSelector.tsx
```diff
@@ -0,0 +1,211 @@
...
+  setFilterOptions,
+}: Pick<THeaderProps, 'types' | 'setFilterOptions'>) => {
+  const nexus = useNexusContext();
+  const originTypes = useRef<TType[]>([]);
+  const [typeSearchValue, updateSearchType] = useState('');
+  const [typesOptionsArray, setTypesOptionsArray] = useState<TType[]>([]);
+
+  const selectCallback = useCallback((data: any) => {
+    const options = (
+      ((data as unknown) as TTypeAggregationsResult).aggregations.types
```
Dinika, line 98: imho we should do this translation of types (`data as unknown as TTypeAggregationsResult` in the `useAggregations` hook. The user of the this hook should be able to safely guess the type as `TTypeAggregationsResult` from the value returned by the hook:

```javascript
const useTypesAggregation = ({
  nexus,
  selectCallback,
}: SomeType) => {
  return useQuery({
    ...
    select: (data) => selectCallback(data as unknown as TTypesAggregatedBucket),
  });
};

// user of the hook
  const selectCallback = useCallback((data: TTypeAggregationsResult) => {
    const options = (
    data.aggregations.types
        ?.buckets ?? ([] as TTypesAggregatedBucket[])
    )
    ...
  }, []);
    const { data: typeOptions } = useTypesAggregation({
    nexus,
    selectCallback,
  });
```

Dinika (approved):

Dinika (approved):
Looks good but tests are missing.

src/shared/canvas/MyData/MyData.tsx
bilalesi: it's not always there, in the query, if you want to not include some fields in the `queryParams` you have to make them `undefined`, and this is why the types must be `undefined`

PR #1340, opened 2023-07-18
US 4041 // Collapse header when scrolled away

Dinika:
https://github.com/BlueBrain/nexus-web/assets/11242410/f70f267b-aed0-4c5a-a91a-d04d7f956d29

## Types of changes

## Checklist:


src/subapps/dataExplorer/DataExplorerCollapsibleHeader.tsx
```diff
@@ -0,0 +1,129 @@
...
+  useLayoutEffect(() => {
+    const handleScroll = () => {
+      if (throttleTimeout === null) {
+        throttleTimeout = window.setTimeout(callBack, waitMs);
+      }
+    };
+
+    window.addEventListener('scroll', handleScroll);
+
+    return () => window.removeEventListener('scroll', handleScroll);
```
bilalesi, line 127: need to clear the `timeout` here in the cleanup fn of the `useLayoutEffect`
setting the throttleTimeout to null will not be sufficient, the timeout event will keep running until you cleared it
clearing the timeout will avoid  unpredictable behaviours and memory leaks.

src/subapps/dataExplorer/DataExplorerCollapsibleHeader.tsx
```diff
@@ -0,0 +1,134 @@
...
+) {
+  const yPosition = useRef(getScrollYPosition());
+
+  let throttleTimeout: number | null = null;
+
+  const callBack = () => {
+    const currentPosition = getScrollYPosition();
+    effect(currentPosition);
+    yPosition.current = currentPosition;
+    throttleTimeout = null;
```
bilalesi, line 115: since you are setting the `throttleTimeout` to null here, then the clearing will not gonna happen at all; so what we can do is :
remove the `throttleTimeout = null` from the callback, the `setTimeout` will create one when you call the fn ,
declare the `throttleTimeout` inside the `useLayoutEffect` so the cleanup can access it

bilalesi (approved):

src/subapps/dataExplorer/DataExplorerCollapsibleHeader.tsx
```diff
@@ -106,29 +107,20 @@ export function useScrollPosition(
...
-      if (throttleTimeout === null) {
-        throttleTimeout = window.setTimeout(callBack, waitMs);
-      }
-    };
-
-    window.addEventListener('scroll', handleScroll);
+    const throttledCallback = throttle(callBack, waitMs, {
+      leading: true,
+    });
+    window.addEventListener('scroll', throttledCallback);
```
bilalesi, line 120: yeah nice one 👍

bilalesi (approved):

PR #1341, opened 2023-07-19
Hotfix/slow navigation in expand

bilalesi:
Fixes #

## Description
1. add a redux middleware listner that capture every eligible actions to save the navigation history in `sessionStorage`
and calculate the digest.
2. remove every tooltip when the user click on a link

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/components/ResourceEditor/useEditorTooltip.tsx
+++ b/src/shared/components/ResourceEditor/useEditorTooltip.tsx
@@ -219,7 +219,7 @@ function useEditorTooltip({
           tooltip.remove();
         }
         return clearTimeout(timeoutId);
-      }, 2000);
+      }, 3000);
 
       return tooltip;
     }
--- a/src/shared/components/ResourceEditor/useEditorTooltip.tsx
+++ b/src/shared/components/ResourceEditor/useEditorTooltip.tsx
@@ -105,8 +105,9 @@ function createTooltipContent({ resolvedAs, error, results }: TTooltipCreator) {
     tooltipContent.appendChild(
       createTooltipNode({
         tag: 'Multiple',
-        title: `${(results as TDELink[]).length
-          } resources was found, click to list them`,
+        title: `${
+          (results as TDELink[]).length
+        } resources was found, click to list them`,
       })
     );
     return tooltipContent;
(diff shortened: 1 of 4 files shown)


src/shared/components/ResourceEditor/useEditorTooltip.tsx
```diff
@@ -223,7 +219,7 @@ function useEditorTooltip({
           tooltip.remove();
         }
         return clearTimeout(timeoutId);
-      }, 3000);
+      }, 2000);
```
Dinika, line 222: If the timeout is for what I think it is (I think it is to hide the tooltip that is shown when user hovers on link), then I think 3 seconds might be a better timeout than 2 because it would give more time to the user to read the contents of the tooltip.

src/shared/store/reducers/data-explorer.ts
```diff
@@ -196,7 +200,7 @@ export const dataExplorerSlice = createSlice({
           shrinked: isShrinkable(rightNodesLinks),
         },
       };
-      calculateNewDigest(newState);
+      // calculateNewDigest(newState);
```
Dinika, line 203: could we remove this completely?

src/shared/store/reducers/data-explorer.ts
```diff
@@ -79,13 +83,13 @@ const calculateNewDigest = (state: TDataExplorerState) => {
 const isShrinkable = (links: TDELink[]) => {
   return links.length > MAX_NAVIGATION_ITEMS_IN_STACK;
 };
-function insert(arr: any[], index: number, item: any) {
-  arr.splice(index, 0, item);
-}
+
+const DataExplorerFlowSliceName = 'data-explorer-graph-flow';
+const DataExplorerFlowSliceListener = createListenerMiddleware();
```
Dinika, line 88: Really cool. I never used this before.

Dinika (approved):

src/shared/components/ResourceEditor/useEditorTooltip.tsx
bilalesi: yeah completely agree, I was just doing an experiment and forget about it

PR #1361, opened 2023-07-31
F4096: Multi type selector

bilalesi:
Fixes #https://github.com/BlueBrain/nexus/issues/4096

## Types of changes

## Checklist:


src/shared/molecules/TypeSelector/TypeSelector.tsx
```diff
@@ -0,0 +1,232 @@
+import React, { useCallback, useRef, useState, ReactElement } from 'react';
```
Dinika, line 1: Nice to have: I think the file hierarchy is a bit weird. Instead of placing this file inside `src/shared/molecules`, it should probably be in `src/shared/components`.

The My-Data stuff could go in `src/containers/MyData` or `src/subapps/MyData`. Same for DataExplorer too.

src/subapps/dataExplorer/ColumnsSelector.tsx
```diff
@@ -0,0 +1,123 @@
...
+    {
+      inputColumns: columns,
+      queryTerm: '',
+    }
+  );
+  const onSearchValueChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (e.type === 'click' || e.target.value === '') {
+      updateSearchResults({ inputColumns: columns, queryTerm: '' });
+    }
+    updateSearchResults({
```
Dinika, line 55: I think we should add a `return;` in the `if` condition above. Otherwise the `updateSearchResults` will be called twice if the condition is true.


Also, I'm not sure I understand the purpose of `if` block here. Why do we want to set `inputColumns` as all the `columns`? Shouldn't `inputColumns` be set to only those columns where the value matches the value of the clicked item?

src/subapps/dataExplorer/ColumnsSelector.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }, [columns]);
+
+  return (
+    <div className="columns-selector">
+      <div className="columns-selector-wrapper">
+        <label htmlFor="columns-selector">Columns: </label>
+        <Select
+          id="columns-selector"
+          style={{ width: 300 }}
+          placeholder="custom dropdown render"
```
Dinika, line 73: The `placeholder` in `Select` item is not needed.

src/subapps/dataExplorer/ColumnsSelector.tsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+  );
+  const onSearchValueChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    if (e.type === 'click' || e.target.value === '') {
+      updateSearchResults({ inputColumns: columns, queryTerm: '' });
+    }
+    updateSearchResults({
+      queryTerm: e.target.value,
+      inputColumns: columns.filter(column =>
+        column.value.toLowerCase().includes(e.target.value.toLowerCase())
```
Dinika, line 58: There's a `normalizeString` function in `src/utils/stringUtils.ts` that can be used here for normalizing both, `column.value` and `e.target.value`

Dinika (approved):
Looks good overall. [This comment] is important. Unit tests for the column selector are missing.

PR #1379, opened 2023-08-18
4010 // Update query to retrieve non-deprecated resources by default

Dinika:
Fixes #[4010](https://app.zenhub.com/workspaces/blue-brain-nexus-60ace035034cae00105c9307/issues/gh/bluebrain/nexus/4010)
Shows non-deprecated resources by default

*Note*
Since the default query (which was faulty until now) is saved in the local storage the users will need to clear their local storage to see the fixed result. If needed, I can also update this PR to override the saved query retrieved from the LS, with `deprecated: false`. I'm open to suggestions. 

## Types of changes

## Checklist:


bilalesi:
there is an issue with the total count of results
when the user select a type it will show the total of deprecated + un-deprecated 
then if the user play with deprecated toggle on/off, 

video send in slack

bilalesi (approved):

PR #1388, opened 2023-09-18
Fusion update fix tests

Dinika:
Fixes #

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM timbru31/node-alpine-git:16 as builder
+FROM timbru31/node-alpine-git as builder
 
 WORKDIR /tmp/nexus-web
 COPY . /tmp/nexus-web
 ENV GENERATE_SOURCEMAP=false
```
bilalesi, line 5: we can remove `GENERATE_SOURCEMAP`, vite don't use this env

Dockerfile
```diff
@@ -1,11 +1,11 @@
...
 RUN yarn && yarn --max-old-space-size=6144 build
 
-FROM node:16-alpine
+FROM node:18-alpine
 WORKDIR /opt/nexus
 COPY --from=builder /tmp/nexus-web/dist /opt/nexus
 EXPOSE 8000
```
bilalesi, line 11: we can also update the entrypoint since we have the `preview` script to
`ENTRYPOINT ["npm", "run", "preview"]` since we don't have yarn installed by default in node alpine (to confirm)

bilalesi:
nice job 👍

Dockerfile
Dinika: Really good point. I'll remove this.

PR #1400, opened 2023-10-04
Hotfix/update build configuration

bilalesi:
Fixes #

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -5,7 +5,8 @@ COPY . /tmp/nexus-web
 RUN yarn && yarn --max-old-space-size=6144 build
 
 FROM node:18-alpine
+ENV NODE_ENV=production
 WORKDIR /opt/nexus
-COPY --from=builder /tmp/nexus-web/dist /opt/nexus
+COPY --from=builder /tmp/nexus-web/dist /opt/nexus/dist
```
bilalesi, line 10: I added /dist to the path; so we do not all other config to the server to get the static assets

etsc.config.js
```diff
@@ -0,0 +1,16 @@
...
+        target: "es2020",
+        outdir: './dist/server',
+        bundle: true,
+        loader: { ".node": "file" },
+        external: [
+            "esbuild",
+            "lightningcss"
+        ]
+    },
+};
```
bilalesi, line 16: The idea is the following:
there is two ways how we can start the server in production env (docker)
1. either include the `node_modules` in the docker image so we can use `ts-node` to start the server, 
2. Build the server app and bundle it so we did not need the `node_modules`
 I did not find a good way to build the server with `Vite`, so I use `esbuild` which is also used with `Vite` in dev mode

package.json
```diff
@@ -7,11 +7,15 @@
...
-    "build": "rm -rf ./dist && vite build -c ./vite.config.refresh.js && vite build",
-    "preview": "cross-env NODE_ENV=production node server",
+    "clean": "rm -rf ./dist",
+    "dev:client": "yarn clean && concurrently 'vite build -c ./vite.config.refresh.js' 'vite'",
+    "dev:server": "NODE_ENV=development ts-node-esm --project tsconfig.node.json -- ./server/index.ts",
+    "dev": "concurrently 'yarn dev:client' 'yarn dev:server'",
+    "build:server": "etsc --config=etsc.config.js",
+    "build:client": "vite build -c ./vite.config.refresh.js && vite build",
+    "build": "yarn clean && yarn build:client && yarn build:server",
+    "preview": "cross-env NODE_ENV=production node ./dist/server",
```
bilalesi, line 18: Update dev and build scripts
in dev: using `ts-node` for server
in prod: using `esbuild` (estc) for server

server/index.ts
```diff
@@ -10,19 +9,19 @@ import {
...
-
+} from './constants';
 
-process.env["NODE_TLS_REJECT_UNAUTHORIZED"] = 0;
 const NODE_ENV = process.env.NODE_ENV;
-const PORT = process.env.PORT || 8000;
+const DISABLE_SSL = process.env.DISABLE_SSL;
+const PORT = Number(process.env.PORT) || 8000;
+// @ts-ignore
+process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = DISABLE_SSL;
```
bilalesi, line 18: meant to be only for dev mode, wondering why an error raised due SSL cert, will investigate it (hopefully will not be in prod)

Dinika (approved):
Two small things:

1. We should probably add in the new readme that we need to pass the `DISABLE_SSL` variable.

2. It might help to add [environment files] (https://vitejs.dev/guide/env-and-mode.html#env-files)  (especially env.dev) to easily start server locally. We can do it later though, just wanted to suggest it.

PR #1402, opened 2023-10-05
Fix/no cert in server

bilalesi:
Move fetching the realms to the browser instead of pre-fetching (even i prefer this way), so to not do more configuration on the deployment side (with K8s but a thing to be consider later), this way will be no more external call in the server that need ssl checking

## Types of changes

## Checklist:


Dinika (commented):
Sounds good. What changes do you mean we'll need to do on k8s side?

Dinika (approved):

bilalesi:
We had to added cert to the configuration and this need a collaboration from the CS team, and it's not a mandatory step for the moment, we need just one request to added in the server (fetching the realms to be pre-rendered with the first HTML)

PR #1403, opened 2023-10-06
Fix/serve static files

bilalesi:
1. Add base path to history browser router
2. change configuration of the server to be built inside the dist folder so we can use `__dirname` for static files
3. change configuration of the client build.
4. Add request logger to the server

## Types of changes

## Checklist:


index.html
```diff
@@ -2,7 +2,6 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
```
Dinika, line 5: nice catch

server/index.ts
```diff
@@ -104,7 +100,7 @@ async function transformer(html: string, req: Request) {
     .replace(
       '<!--app-state-->',
       `<script>
-        window.__BASE__ = '/';
+        window.__BASE__ = '${base}';
```
Dinika, line 103: Bit confused about this. Here we are replacing a string with another string which will have the content as `'${base}'`. Could this be a mistake? The string interpolation is not enclosed with backticks (``), they are within `''` and it's not clear to me how and when the value will be changed to right value ('/'). Are we sure it shouldn't be `window.__BASE__ = '/';`?

Dinika (approved):

server/index.ts
bilalesi: so the string interpolation is already wrapping all the script `line 102`, and to show the base inside the `dom` as string we should wrap it with the quotes '' as `window.__BASE__ ='/nexus'`, unless it will raise an error as `window.__BASE__ = /nexus`
for why we should include the base, because if you look to the ingress we have prefixes for each env, for staging for example`/nexus` for prod we have `/nexus/web/`.

PR #1407, opened 2023-10-12
F4349/id resolution page

bilalesi:
Fixes #https://github.com/BlueBrain/nexus/issues/4349

**1. User has access to the resource, user is authenticated, and the resolution result is one single resource**
-  User access `{fusionBaseUrl}/resolve/{encodedResourceId}`.
-  The user will be redirected to the resource page.

**2. User has access to the resource, the user is not authenticated, and the the resolution result is one single resource**
-  User access `{fusionBaseUrl}/resolve/{encodedResourceId}`.
-  The user will be redirected to the login page.
-  If the user successfully authenticated will be redirected to the resource page.

**3. User has not access to the resource**
-  User access `{fusionBaseUrl}/resolve/{encodedResourceId}`.
- A modal will be opened to the user to ask him to logout and choose another realm

**4. User has access to the resource, and the resolution result is `NoneOrMany` resources**
> _(Note: This is part not yet can be tested due Delta not support it at this moment)_
> _(Note: The design may change in the future, the current one just for test)_
- User access `{fusionBaseUrl}/resolve/{encodedResourceId}`.
- The user will be redirected to the login page if not authenticated.
- If the user successfully authenticated will be redirected to the `/resolved/{encodedResourceId}`.
- If empty result -> A simple message 'No result found'
- If Many resources has been resolved, A collapsible Item with different resources 

## Types of changes

## Checklist:

### DIFF
--- a/src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
+++ b/src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
@@ -43,7 +43,7 @@ const IDResolveRedirectionPage = () => {
 
   const { error, isError, isLoading } = useQuery({
     enabled: isAuthenticated,
-    queryKey: ['resolver', { apiEndpoint, resourceId }],
+    queryKey: ['resource-id-resolver', { apiEndpoint, resourceId }],
     queryFn: () =>
       nexus.httpGet({
         path: `${apiEndpoint}/resolve/${resourceId}`,
@@ -54,7 +54,6 @@ const IDResolveRedirectionPage = () => {
     onSuccess: data => console.log('@@dataResolution', data),
     onError: error => console.log('@@errorResolution', error),
   });
-  console.log('@@isError', isError, error, JSON.stringify(error, null, 2));
 
   const handleReconnection = () => {
     localStorage.removeItem('nexus__state');
--- a/src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
+++ b/src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
@@ -1,0 +1,105 @@
+import { CSSProperties } from 'react';
+import { useQuery } from 'react-query';
+import { useSelector } from 'react-redux';
+import { useHistory, useParams } from 'react-router';
+import { useNexusContext } from '@bbp/react-nexus';
+import { Button, Modal } from 'antd';
+import { isEmpty } from 'lodash';
+
+import { RootState } from '../../shared/store/reducers';
+import { ResourceJSONPrettify } from './IDResolvedManyPage';
+
+const modalStyle = {
+  display: 'flex',
+  alignItems: 'flex-start',
+  justifyContent: 'center',
+  flexDirection: 'column',
+} as CSSProperties;
+
+const calloutStyle = {
+  margin: 10,
+  width: '100%',
+  border: '1px solid #c11b1b3b',
+  padding: '4px',
+  borderRadius: '4px',
+  fontSize: 14,
+  fontWeight: 'bold',
+} as CSSProperties;
+
+const IDResolveRedirectionPage = () => {
+  const nexus = useNexusContext();
+  const oidc = useSelector((state: RootState) => state.oidc);
+
+  const { push: navigate } = useHistory();
+  const { apiEndpoint } = useSelector((state: RootState) => state.config);
+  const { resourceId } = useParams<{ resourceId: string }>();
+
+  const token = oidc && oidc.user && !!oidc.user.access_token;
+  const authenticated = !isEmpty(oidc.user);
+  const isAuthenticated = authenticated && token;
+
+  // we should encode it again due oidc returning the url not encoded
+  const redirectUri = `/resolve/${encodeURIComponent(resourceId)}`;
+
+  const { error, isError, isLoading } = useQuery({
+    enabled: isAuthenticated,
+    queryKey: ['resolver', { apiEndpoint, resourceId }],
+    queryFn: () =>
+      nexus.httpGet({
+        path: `${apiEndpoint}/resolve/${resourceId}`,
+        // headers: { Accept: 'text/html' }
+      }),
+    retry: false,
+    refetchOnWindowFocus: false,
+    onSuccess: data => console.log('@@dataResolution', data),
+    onError: error => console.log('@@errorResolution', error),
+  });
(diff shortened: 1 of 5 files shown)


src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -0,0 +1,104 @@
...
+  border: '1px solid #c11b1b3b',
+  padding: '4px',
+  borderRadius: '4px',
+  fontSize: 14,
+  fontWeight: 'bold',
+} as CSSProperties;
+
+const IDResolveRedirectionPage = () => {
+  const nexus = useNexusContext();
+  const oidc = useSelector((state: RootState) => state.oidc);
```
Dinika, line 31: I think it may make sense to have the selector directly return if the user is authenticated or not. (Maybe we can even have that as a global selector if possible). Something like this:

```typescript
const isAuthenticated = useSelector((state) => {
    const oidc = state.oidc;
    const token = oidc && oidc.user && !!oidc.user.access_token;
    return !isEmpty(oidc.user) && token;
});
```
What do you think?

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -0,0 +1,104 @@
...
+  // we should encode it again due oidc returning the url not encoded
+  const redirectUri = `/resolve/${encodeURIComponent(resourceId)}`;
+
+  const { error, isError, isLoading } = useQuery({
+    enabled: isAuthenticated,
+    queryKey: ['resource-id-resolver', { apiEndpoint, resourceId }],
+    queryFn: () =>
+      nexus.httpGet({
+        path: `${apiEndpoint}/resolve/${resourceId}`,
+        // headers: { Accept: 'text/html' }
```
Dinika, line 50: Is this comment needed?

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -0,0 +1,104 @@
...
+    onSuccess: data => console.log('@@dataResolution', data),
+    onError: error => console.log('@@errorResolution', error),
+  });
+
+  const handleReconnection = () => {
+    localStorage.removeItem('nexus__state');
+    navigate(`/login?destination=${redirectUri}`);
+  };
+
+  if (!isLoading && isError) {
```
Dinika, line 63: It's not clear to me why we need the `!isLoading` part. Shouldn't just `if (isError)` be enough? What am i missing?

Dinika (commented):
Looks good to me. Testing now.

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -0,0 +1,104 @@
+import { CSSProperties } from 'react';
```
Dinika, line 1: @bilalesi 

I'm having some troubles testing.
I have started fusion locally. I'm using delta dev endpoint and I am logged in as localuser.

I am trying to resolve this resource which (according to the metadata) is created by localuser:
https://dev.nise.bbp.epfl.ch/nexus/v1/resources/copies/hippocampus-hub/_/TestResource2

I then try to go to this url, but I see an empty page. (I have encoded the `@id` of the resource)

http://localhost:8000/resolve/https%3A%2F%2Fdev.nise.bbp.epfl.ch%2Fnexus%2Fv1%2Fresources%2Fcopies%2Fhippocampus-hub%2F_%2FTestResource2

Screenshot from 2023-10-12 17-35-10

What am I doing wrong?

Dinika: When I encode the `self` of the resource, I see an error. I was actually expecting to see the resolved resource
Screenshot from 2023-10-12 17-37-30

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
bilalesi: Actually I have to uncomment it, that the important thing

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const nexus = useNexusContext();
+  const oidc = useSelector((state: RootState) => state.oidc);
+
+  const { push: navigate } = useHistory();
+  const { apiEndpoint } = useSelector((state: RootState) => state.config);
+  const { resourceId } = useParams<{ resourceId: string }>();
+
+  const token = oidc && oidc.user && !!oidc.user.access_token;
+  const authenticated = !isEmpty(oidc.user);
+  const isAuthenticated = authenticated && token;
```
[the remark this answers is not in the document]
bilalesi: yeah true, was just copy past, did not focus on it, will change it

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
bilalesi: we can do that, but we should `memoize` the `selector fn`

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
bilalesi: the`ID != self`, so will not work, delta resolve endpoint was built for IDs, we already have this functionality for self working

bilalesi: I fixed the issue, I commented out the accept header, this is why you are seeing the empty page
In the fix, you will get this message: (we can refine the text later) due cors

Screenshot 2023-10-12 at 19 31 15

Dinika (approved):

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -0,0 +1,105 @@
+import { CSSProperties, useCallback, useMemo } from 'react';
```
Dinika, line 1: @bilalesi 
Looks good. My only suggestion is to change the error message from "The redirection performance is suboptimal; further investigation is required, Please check again", to something like "The redirection feature is in beta version. We have taken note of this error and are working on improving the feature".

This is not mandatory though and very subjective.

Dinika (approved):

PR #1409, opened 2023-10-16
fix: redirect manually instead of redirect-follow

bilalesi:
Fixes #

## Types of changes

## Checklist:

### DIFF
--- a/src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
+++ b/src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
@@ -1,8 +1,6 @@
-import { CSSProperties, useCallback, useMemo } from 'react';
-import { useQuery } from 'react-query';
+import { useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { useHistory, useParams } from 'react-router';
-import { useNexusContext } from '@bbp/react-nexus';
 import { Button, Modal } from 'antd';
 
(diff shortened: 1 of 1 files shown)


src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -14,12 +12,15 @@ import {
 } from './IDResolvedManyPage';
 
 const IDResolveRedirectionPage = () => {
-  const nexus = useNexusContext();
   const navigate = useHistory().push;
   const { apiEndpoint, basePath } = useSelector((state: RootState) => ({
     apiEndpoint: state.config.apiEndpoint,
     basePath: state.config.basePath,
   }));
+  const [{ error, isError }, setResolutionState] = useState({
```
Dinika, line 20: Two things about this state:

1. I think the name `setResolutionError` or ``setResourceResolutionError` is more suited here because we only store error information in this state, nothing else
2. Where are we using the value of `error` state? Should we perhaps add a notification displaying this error?

Dinika (approved):
Left some comments but if you want you can deploy and test the PR and resolve the comments later based on the test results.

src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
bilalesi: 1. 👍 
2. the error will be showed in the modal

PR #1411, opened 2023-10-16
fix: use replace to redirect

bilalesi:
the fetch specification mentioned that when redirection happening we should get `redirected: true` but i actually not the case for us, so removing the check and using `replace` in the history navigation fix the issue

## Types of changes

## Checklist:


src/pages/IDResolutionPage/IDResolveRedirectionPage.tsx
```diff
@@ -43,18 +44,18 @@ const IDResolveRedirectionPage = () => {
...
           },
           redirect: 'manual',
+          signal: ACInstance.signal,
         })
           .then(res => {
-            setResolutionError(prev => ({ ...prev, isError: false }));
-            if (res.redirected) {
-              window.location.href = res.url;
-            }
+            window.location.replace(res.url);
```
Dinika, line 50: In this case, shouldn't we set `resolutionError.isError` to false?

Dinika (approved):

PR #1419, opened 2023-10-23
Migration: docker image

bilalesi:
create a new docker image for testing migration in dev cluster

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/migrate.yml
+++ b/.github/workflows/migrate.yml
@@ -1,0 +1,24 @@
+name: Migration
+on:
+  push:
+    branches:
+      - fusion-react-18-vite
+
+jobs:
+  publish-preview:
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build
+        run: |
+          docker build -t bluebrain/nexus-web:migrate .
+      - name: Publish To DockerHub
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/migrate.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  push:
+    branches:
+      - fusion-react-18-vite
+
+jobs:
+  publish-preview:
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
```
Dinika, line 11: Is there a reason why we've added the timeout. Nothing wrong with, just curious.

src/shared/organisms/DataPanel/DataPanel.spec.ts
```diff
@@ -1,14 +1,13 @@
-import { vi } from 'vitest';
+import { vi as jest } from 'vitest';
```
Dinika, line 1: why this change? Maybe it's better to rename the occurences of the word `jest` underneath to `vitest`.

Dinika (changes requested):

.github/workflows/migrate.yml
bilalesi: this is was the same configuration in preview, i did not may attention but it won't make any issue

Dinika (approved):

src/shared/organisms/DataPanel/DataPanel.spec.ts
bilalesi: you are right about it i fixed it, in the next PR

PR #1424, opened 2023-10-30
Add diff fields depends on the added/deleted/updated status

bilalesi:
1. fix the json to show the values with undefined using the replaced param of `JSON.stringify`
2. fix the style of the diffing object using `<pre/> tag`
3. return the diffing object with three different sub-categories of the diffing (`added/deleted/updated`)
## Description

Screenshot 2023-10-31 at 08 34 10

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/components/History/index.tsx
+++ b/src/shared/components/History/index.tsx
@@ -34,6 +34,7 @@ const HistoryComponent: React.FunctionComponent<{
             </Timeline.Item>
           );
         }
+
         return (
           <Timeline.Item
             key={index}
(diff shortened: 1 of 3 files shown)


src/shared/components/History/index.tsx
```diff
@@ -45,10 +46,20 @@ const HistoryComponent: React.FunctionComponent<{
...
+                style={{ width: '100%', marginTop: '1em' }}
               >
                 {revision.hasChanges ? (
-                  <Card>{JSON.stringify(revision.changes, null, 2)}</Card>
+                  <Card>
+                    <pre style={{ width: '100%' }}>
+                      {JSON.stringify(
+                        revision.changes,
+                        (key, value) => {
+                          return typeof value === 'undefined' ? null : value;
```
Dinika, line 57: If it is the case that `value` can be null also then it may make sense to protect against that too (maybe using isNil from lodash). Otherwise this is good.

Dinika (approved):

src/shared/components/History/index.tsx
bilalesi: here, we are checking if the key/value is exist in the object, basically the diff method, will assign the value undefined when it can not find it in the object, in the resource objects we will not find the null value, since `null` it's not available in python (mostly used here).

PR #1432, opened 2023-11-07
Fix: CodeEditor duplication in Dev mode

bilalesi:
Code editor is render twice in develop mode due react 18 state restoration feature
In Build version everything will work just fine

## Types of changes

## Checklist:

### DIFF
--- a/src/shared/components/ResourceEditor/CodeEditor.tsx
+++ b/src/shared/components/ResourceEditor/CodeEditor.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useRef } from 'react';
 import { EditorConfiguration } from 'codemirror';
 import { UnControlled as CodeMirror } from 'react-codemirror2';
 import { INDENT_UNIT, highlightUrlOverlay } from './editorUtils';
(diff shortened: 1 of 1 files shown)


src/shared/components/ResourceEditor/CodeEditor.tsx
```diff
@@ -53,11 +54,20 @@ const CodeEditor = forwardRef<CodeMirror.Editor | undefined, TCodeEditor>(
...
             highlightUrlOverlay(editor);
             (ref as React.MutableRefObject<CodeMirror.Editor>).current = editor;
           }}
+          editorWillUnmount={() => {
+            const editor = (ref as React.MutableRefObject<
+              CodeMirror.Editor
+            >).current.getWrapperElement();
+            if (editor) editor.remove();
+            if (wrapperRef.current)
+              (wrapperRef.current as { hydrated: boolean }).hydrated = false;
```
Dinika, line 69: I'm not sure I understand what is going on in line 69. Could we add a comment? It would definitely help me.
Not a must have though.

Dinika (approved):

src/shared/components/ResourceEditor/CodeEditor.tsx
bilalesi: surely, yeah i can explain,
If you check the source code there is a function called `hydarte` and it's responsible to do state preservation resetting in mount and update and in this function we set the value of `hydrated` to true (by the way the hydrated value is private property not exposed) so we have to set it to false after unmount so the next render we initiate the editor in the hydrate fn 
```
if (!this.hydrated) {
      const doc = this.editor.getDoc();
      const lastLine = doc.lastLine();
      const lastChar = doc.getLine(doc.lastLine()).length;

      doc.replaceRange(props.value || '',
        {line: 0, ch: 0},
        {line: lastLine, ch: lastChar});
}
```

if you try to remove the `hydrated = false` in the un-mounting, then in the next rendering the modified value will not preserved 


hope i explain it well

PR #1438, opened 2023-11-27
Undeprecate resources

danburonline:
Fixes [#4509](https://github.com/BlueBrain/nexus/issues/4509)

### Major things

- Added an undo deprecate button to the `ResourceView` component. When clicking on it it updates the states and goes to the latest revision (it's the same flow as when you, e.g. deprecate a resource).
deprecated-undo
- Don't show the undo deprecation button if it's an older resource revision.
- Added a toast/notification in case the un-deprecation failed
image
- Don't allow undoing deprecation for unsupported resource types. Example:
CleanShot 2023-11-30 at 11  06 15. This would need to be changed as soon as we allow the deprecation of other resource types as well.
- Hide the button to deprecate the default Elasticsearch view, because it would break the listing operations (as described by @imsdu):
no-deprecate-button

### Cleaning

- Sorted the `package.json` via [`sort-package-json`](https://www.npmjs.com/package/sort-package-json) => following common rules.
- - Unit tests failed after my adjustments as the `QueryClient` was missing. This is fixed.
- Sorted imports of files I worked at. The ESLint rule is still missing (or generally ESLint).
- Replaced the bitmap logo with an SVG.
- Worked on reducing the delta from design to code by adjusting the logo size and the main white spaces on the pages. See the screenshot below for a before and after:

before-after

### Additional Tickets Uncovered

- [4541](https://github.com/BlueBrain/nexus/issues/4541): The test coverage fell due to my adjustments. But to add new tests for the `ResourceViewContainer` component. This was out of scope for the [the current ticket of this PR (#4509)](https://github.com/BlueBrain/nexus/issues/4509)
- [4542](https://github.com/BlueBrain/nexus/issues/4542): During the development I realised that the UI is quite messy, as it potentially grew organically (Frankenstein UI). This new ticket discusses a potential non-functional redesign.
- [4543](https://github.com/BlueBrain/nexus/issues/4543): During the development, I realised that the copy of the app was not consistent and, in general, maybe sometimes too technical/not conforming the docs and/or the mental model of Nexus. This new ticket discusses this.

## How has this been tested?

- Current tests pass.
- Tested the flows locally via prod build.
- I added an E2E test that tests the happy-path flow of deprecating a deprecatable resource. If I should add more E2E tests, let me know. I think the other non-happy paths can be tested via unit tests (as mentioned in [4541](https://github.com/BlueBrain/nexus/issues/4541)).

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -608,14 +606,64 @@ const ResourceViewContainer: React.FunctionComponent<{
...
+            orgLabel,
+            projectLabel
+          ),
+        });
+
+        setLatestResource({
+          ...latestResource!,
+          _rev: latestResource!._rev + 1,
+          _deprecated: false,
+        });
```
bilalesi, line 652: no need to set the state since there is a redirection just after

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -608,14 +606,64 @@ const ResourceViewContainer: React.FunctionComponent<{
...
+
+        goToResource(orgLabel, projectLabel, resourceId, {
+          revision: latestResource!._rev + 1, // Go to the n + 1 = latest revision after the un-deprecation
+        });
```
bilalesi, line 656: no. need to specify the revision, when not specifying the revision the new page will fetch the latest revision by default

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -726,10 +774,53 @@ const ResourceViewContainer: React.FunctionComponent<{
...
+                                  <Button
+                                    icon={<UndoOutlined />}
+                                    style={{
+                                      marginTop: '10px',
+                                      marginBottom: '5px',
+                                    }}
+                                    onClick={async () => {
+                                      unDeprecateResource();
+                                    }}
+                                  >
```
bilalesi, line 799: no need for async here, just pass the fn to on click
`onClick={() => unDeprecateResource()}`

src/shared/organisms/DataExplorerGraphFlowContent/DataExplorerGraphFlowContent.tsx
```diff
@@ -13,7 +13,7 @@ const DataExplorerContentPage = ({}) => {
   return (
     <div className="degf-content__wrapper">
       {fullscreen ? (
-        <Fragment>
+        <>
```
bilalesi, line 16: for `Fragment` vs `<></>`, it's a personal preference 
one benefit of Fragment is that you can use keys  when you need it as mapping a list

bilalesi (commented):
good work, left some comments

regarding this part: 
https://github.com/BlueBrain/nexus/issues/4542: During the development I realised that the UI is quite messy, as it potentially grew organically (Frankenstein UI). This new ticket discusses a potential non-functional redesign.


There is already a design for the Resource details view and we already have a ticket for that 
I will share it with you when we get there

src/shared/organisms/DataExplorerGraphFlowContent/DataExplorerGraphFlowContent.tsx
danburonline: It's true. If no mapping is needed, then I think the shorthand looks cleaner as it implies that there won't be a DOM node added in the rendered HTML.

src/shared/containers/ResourceViewContainer.tsx
danburonline: Good point. I double-checked and indeed it doesn't make a difference other than not letting React decide to update the state. However, see my reply on your other comment.

src/shared/containers/ResourceViewContainer.tsx
danburonline: I quickly re-tested it, and I found out that this is not true. 
image

When not performing this action, the component's state will be updated, but React does not automatically rerender the component. To ensure that it is rendered, I enforce the update. Since it is a state based on the route, it is preferable to increment it by 1 and truly force the update. I could adjust the helper function or just do a +1, my thoughts were that the +1 is cleaner.

src/shared/containers/ResourceViewContainer.tsx
danburonline: Good catch! Fixed it.

src/shared/organisms/DataExplorerGraphFlowContent/DataExplorerGraphFlowContent.tsx
bilalesi: That's what I said so basically if someone use the fragment we do not need to rename it (ex: For me fragment is more clean ) it's clear and understandable for all the devs that fragment === <></>

src/shared/containers/ResourceViewContainer.tsx
bilalesi: may be you get me wrong, when redirecting you don't have to set the revision 
meaning:
`goToResource(orgLabel, projectLabel, resourceId, {})`

so in brief, no need to state the state because you are redirecting and no need to specify the revision because the page will fetch the latest by default

danburonline: Ah, indeed, I misunderstood you there! That does make sense. I have fixed it, thanks!

cypress/e2e/Resources.cy.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    // Click the "Advanced View" collapse header
+    cy.get('.ant-collapse-header')
+      .contains('Advanced View')
+      .click();
+
+    // Click the "Deprecate" button
+    cy.get('button')
+      .contains('Deprecate', {
+        timeout: 2000, // Just in case for the popover to appear
```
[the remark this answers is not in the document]
danburonline: Works without it – thanks!

src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -608,14 +606,56 @@ const ResourceViewContainer: React.FunctionComponent<{
...
+    projectLabel: string
+  ) {
+    return `${apiEndpoint}/${
+      resource!['@type'] === 'File' ? 'files' : 'resources'
+    }/${orgLabel}/${projectLabel}/${
+      resource!['@type'] === 'File' ? '' : '_/'
+    }${encodeURIComponent(resource!['@id'])}/undeprecate?rev=${
+      latestResource!._rev
+    }`;
+  }
```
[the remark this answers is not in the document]
danburonline: Done.

PR #1439, opened 2023-12-04
Fix double-extension on file download

danburonline:
Fixes [#4529](https://github.com/BlueBrain/nexus/issues/4529)

## Description

pr

Before, when you downloaded a single file via the interface, let's say `example.json`, the download window would show `example.json.json`, which is incorrect. With this change, this won't happen again.

## How has this been tested?

- Locally with dev and prod builds and manual uploading files with extensions, with no extensions and double extensions, such as `example.json.json`.
- Added unit tests for several test cases.

## Types of changes

## Checklist:


src/shared/utils/download.ts
```diff
@@ -2,13 +2,19 @@ import { fileExtensionFromResourceEncoding } from '../../utils/contentTypes';
 
 export const download = (filename: string, mediaType: string, data: any) => {
   const blob = new Blob([data], { type: mediaType });
-  const extention = fileExtensionFromResourceEncoding(mediaType);
+  const extension = fileExtensionFromResourceEncoding(mediaType);
+
+  // Check if filename already ends with the correct extension
+  const hasCorrectExtension = filename.endsWith(`.${extension}`);
+
```
bilalesi, line 9: I am afraid that some data has no `mediaType` and will have to rely on the `filename.ext` 
can you include that in case

bilalesi (approved):
good job, small comment for covering corner case

PR #1440, opened 2023-12-04
Add VS Code settings (debugger, recommended extensions etc.)

danburonline:
Small PR to not ignore the `.vscode` folder in order to include debugger configs and repo extension recommendations. Reason? Especially for the debugger configurations (as we, e.g., also use SSR), to have a higher incentive actually to use the debugger rather than `console.log`.

## Description

- I also added a .editorconfig file, especially useful for the line endings.
- Other than that I only applied one default setting for this workspace: to format on save.

## How has this been tested?

- Tested the debugger configs for the attachment in Chrome on my Mac, the same with the Node.js debugger config.


Dinika:
> @Dinika Bilal told me that he doesn't mind adding these things. What about you? Curious to hear your opinion.

Looks great to me. The debugging config in source code actually a good idea :)

Dinika (approved):

PR #1441, opened 2023-12-04
Improve slack notification for public studios cron job

Dinika:
According to [this solution to the issue](https://github.com/slackapi/slack-github-action/issues/261#issuecomment-1836435861) on the github repo of the plugin we use for slack-github notificate, the blocks sub field can be ignored if the `SLACK_WEBHOOK_TYPE` environment variable is not set. 
In this PR I also (try to) enable the uploading of recorded cypress video to the cypress cloud.

## Types of changes

## Checklist:

### DIFF
--- a/cypress-public-studios.config.ts
+++ b/cypress-public-studios.config.ts
@@ -1,7 +1,6 @@
 const { defineConfig } = require('cypress');
 
 module.exports = defineConfig({
-  projectId: '1iihco',
   e2e: {
     video: true,
     screenshotOnRunFailure: false,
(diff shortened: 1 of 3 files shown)


cypress-public-studios.config.ts
```diff
@@ -1,6 +1,7 @@
 const { defineConfig } = require('cypress');
 
 module.exports = defineConfig({
+  projectId: '1iihco',
```
danburonline, line 4: One suggestion: Shouldn't this be an environment variable?

danburonline (approved):

PR #1442, opened 2023-12-05
Also hide the deprecation of the `defaultSparqlIndex`

danburonline:
This is as in #1438, but with an additional check to hide the deprecation button for `defaultSparqlIndex`. The additional check on the API side of things that one cannot do this still needs to be implemented in the backend. I also used this branch to try to fix the flaky E2E test. While doing so, I cleaned some smaller issues, dead/commented code etc.

## How has this been tested?

Locally (dev + prod build) with my Mac. Not a big change. Tests pass.


danburonline:
Failing because of the flaky E2E test that I removed in the target's branch. Otherwise, it's good to review.

cypress/support/setupRealmsAndUsers.ts
```diff
@@ -93,8 +100,8 @@ const keycloak = keycloakUrl => {
     };
     //@ts-ignore
     return fetch(realmEndpoint(user.realm), request)
-      .then(response => response.json())
-      .then(data => {
+      .then((response: { json: () => any }) => response.json())
+      .then((data: { access_token: any }) => {
```
bilalesi, line 104: `access_token` is `string`

cypress/support/setupRealmsAndUsers.ts
```diff
@@ -93,8 +100,8 @@ const keycloak = keycloakUrl => {
     };
     //@ts-ignore
     return fetch(realmEndpoint(user.realm), request)
-      .then(response => response.json())
-      .then(data => {
+      .then((response: { json: () => any }) => response.json())
```
bilalesi, line 103: response is of type `Response` no need to specify `json`

cypress/support/setupRealmsAndUsers.ts
```diff
@@ -169,54 +161,64 @@ async function setup(users: { [key: string]: User }) {
 
     console.log('Creating internal realm for use by our Service account');
     await getDeltaRealm(internalRealm.name)
-      .then(async response => {
+      .then(async (response: { status: number }) => {
```
bilalesi, line 164: you just need to type `response` as `Response` and everything will automatic

cypress/support/setupRealmsAndUsers.ts
```diff
@@ -169,54 +161,64 @@ async function setup(users: { [key: string]: User }) {
...
           );
-          await createDeltaRealm(internalRealm).then(response => {
-            if (response.status === 201) {
-              console.log('internal realm successfully created');
-            } else {
-              response.text().then(json => {
-                throw Error(`Error occured creating realm in Delta\n\n${json}`);
-              });
+          await createDeltaRealm(internalRealm).then(
+            (response: { status: number; text: () => Promise<any> }) => {
```
bilalesi, line 172: see above 🔝

docs/development.md
```diff
@@ -203,15 +189,15 @@ This means that for development, all we need to run is `ts-node server/index.ts`
 
 ## Lint
 
-We use [ts-lint]() with airbnb rules.
+We use `ts-lint`` with Airbnb rules.
```
bilalesi, line 192: two ticks at the end (see also line 196)

docs/development.md
```diff
@@ -253,28 +239,28 @@ All of the tests will run in headless mode.
 
 ## Debugging End-to-End test failures in CI
 
-There are a few differences between how our e2e tests run in the ci compared to how we run it locally. We use different instances of delta, postgress, keycloak etc in these 2 environments. Also, the version of chrome you may have locally might differ from the version of chrome being used in the ci. Finally, the machine cpu, memory are also most likely different. This can make debugging test failures in ci difficult.
+There are several differences in how our end-to-end tests are executed in the CI compared to local runs. We employ distinct instances of delta, Postgres, Keycloak, and other components in each environment. Moreover, the version of Chrome you have locally may not match the one used in the CI. Additionally, the machine's CPU and memory are likely to vary. These differences can pose challenges when debugging test failures in the CI.
```
bilalesi, line 242: why several and not few, actually few is more correct here

src/shared/containers/ResourceActionsContainer.tsx
```diff
@@ -130,7 +89,7 @@ const ResourceActionsContainer: React.FunctionComponent<{
           message: `Deprecated ${getResourceLabel(resource)}`,
         });
 
-        const { _rev } = deprectatedResource;
+        const { _rev } = deprecatedResource;
```
bilalesi, line 92: if `deprecatedResource` not used elsewhere we can do destruction on top

docs/development.md
danburonline: Changed it back 👍

PR #1443, opened 2023-12-08
Meta tag validation in `CodeEditor` component with custom linter rules

danburonline:
Fixes [#3956](https://github.com/BlueBrain/nexus/issues/3956)

## Description

https://github.com/BlueBrain/nexus-web/assets/34251194/cea7fb8f-eb29-4805-906a-fbdc5337b0d9

This PR introduces a custom linter for validating the JSON-LD format in the Codemirror editor. Currently, I have added one rule: fields cannot start with an underscore. If such fields are found, the linter immediately validates them and displays an error message to the user, including the line number. This change does not impact the generic JSON validation as long as the syntax is correct.

Additionally, I also changed the wording of "Save" to "Save changes" and fixed some minor styling issues, such as padding and the order of buttons (placing the main action on the very right).
## How has this been tested?

- All unit and end-to-end tests have passed.
-  I have added new unit tests to validate all use cases correctly.
-  I have tested it locally on my Mac using both the production and development builds of the app.
-  I have actually worked on optimizing the performance of the `customLinter` by not traversing the entire JSON but only the adjusted lines. I used a large JSON-LD document with over 7000 lines of code. However, as you can see in the commit history, I have reverted these changes as they had a negative impact. I am sticking to the current/original implementation. There is room for improvement, but the overall performance is quite good. Here are the before and after profile screenshots (A = before performance improvements, B = with performance "improvements"):

comparison

Here are the profile files:
- [without-trace.json](https://github.com/BlueBrain/nexus-web/files/13647885/without-trace.json)
- [with-trace.json](https://github.com/BlueBrain/nexus-web/files/13647886/with-trace.json)

## Types of changes

## Checklist:


danburonline:
I still want to do some more profiling to check if I have some memory leaks. But so far it should be ready for a first review.

src/shared/components/ResourceEditor/customLinter.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A custom linter function that checks for issues in a given text.
+ *
+ * The function specifically looks for fields in the text that start with an underscore,
+ * which are considered as errors according to the linter's rules.
+ *
+ * @param text The text to be linted.
+ * @returns An array of LinterIssue objects, each representing a specific issue found in the text.
+ */
```
Dinika, line 19: Love the documentation :star:

src/shared/components/ResourceEditor/index.tsx
```diff
@@ -146,23 +140,41 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
...
   const handleCancel = () => {
     setStringValue(JSON.stringify(rawData, null, 2));
-    setValid(true);
+    setIsValidJSON(true);
     setEditing(false);
   };
 
+  const handleLintError = useCallback(
+    (errors: LinterIssue[]) => {
+      console.log(errors);
```
Dinika, line 149: Is the console.log intentional?

src/shared/components/ResourceEditor/index.tsx
```diff
@@ -173,22 +185,30 @@ const ResourceEditor: React.FunctionComponent<ResourceEditorProps> = props => {
...
+                <div className="feedback _positive">
+                  <CheckCircleOutlined /> Valid
+                </div>
+              )}
+              {!isValidJSON && (
+                <div className="feedback _negative">
+                  <ExclamationCircleOutlined /> Invalid JSON-LD
+                </div>
+              )}
+              {isValidJSON && linterIssues && linterIssues.length > 0 && (
```
Dinika, line 204: Question: According to the type definition `linterIssues` is always an array (right?). If so, why/when do we need to check if it is truthy (I mean this part of the check `..&& linterIssues &&...`)? An array, even if empty, will always be truthy, so do we need this?

The same goes for line 194 above.

Dinika (approved):
Looks great. Love the tests and the performance benchmarking. Leaving small comments and questions that are not blocking.

src/shared/components/ResourceEditor/ResourceEditor.less
```diff
@@ -3,9 +3,12 @@
 @import '../../../../node_modules/codemirror/addon/fold/foldgutter.css';
 @import '../../../../node_modules/codemirror/theme/base16-light.css';
 
-// codemirror styling
-.cm-s-base16-light span.cm-string {
-  // string values
+// CodeMirror styling
```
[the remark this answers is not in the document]
danburonline: I wasn't sure if this was a reference for future editor customization. I simply commented them out so that nothing is selected. But I think we can just remove them.

src/shared/components/ResourceEditor/customLinter.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    const result = customLinter(text);
+    expect(result).toHaveLength(2);
+    expect(result).toContainEqual({
+      message: 'Cannot have fields starting with an underscore',
+      line: 2,
+    });
+    expect(result).toContainEqual({
+      message: 'Cannot have fields starting with an underscore',
+      line: 4,
+    });
```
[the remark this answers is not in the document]
danburonline: Nice one! Thanks. I adjusted it accordingly.

src/shared/components/ResourceEditor/index.tsx
danburonline: No, I'm removing it. (We should consider adding a linter setting to warn about leftover console.logs) 😁

src/shared/components/ResourceEditor/index.tsx
danburonline: That's true. I encountered a linter error when I removed the truthy check. However, I can bypass this issue by initialising the state `const [linterIssues, setLinterIssues] = useState<LinterIssue[]>([]);` with an empty array. Fixed it, thanks!

src/shared/components/ResourceEditor/customLinter.ts
```diff
@@ -0,0 +1,37 @@
...
+  const linterErrors: LinterIssue[] = [];
+  const lines = text.split('\n');
+
+  // Regex to match fields starting with an underscore
+  const regex = /"\s*_(\w+)"/g;
+
+  lines.forEach((line, index) => {
+    if (line.match(regex)) {
+      linterErrors.push({
+        message: 'Cannot have fields starting with an underscore',
```
bilalesi, line 30: I think the message can be improved, like we might explain that underscore fields are reserved for internal use since most of our clients are engineers

danburonline: Fair point. What about: `Fields starting with an underscore are reserved for internal use`?

bilalesi:
can we have the error message in another position, 
moving editor controls seems not that good both UI/UX
may be below the controls (please check with loris if you lost as me)

Screenshot 2023-12-14 at 10 58 58

danburonline:
> can we have the error message in another position, moving editor controls seems not that good both UI/UX may be below the controls (please check with loris if you lost as me)
> 
> Screenshot 2023-12-14 at 10 58 58

Working on it!

danburonline:
> can we have the error message in another position, moving editor controls seems not that good both UI/UX may be below the controls (please check with loris if you lost as me)
> 
> Screenshot 2023-12-14 at 10 58 58

Synced with Loris and adjusted it, have a look:

bilalesi (approved):

PR #1446, opened 2023-12-14
update packages

bilalesi:
This ticket is create due the dependencies warning from `dependabot`

1. Update `express` package
2. remove `qs` querystring package (not needed)

## Types of changes

## Checklist:


package.json
```diff
@@ -51,7 +51,7 @@
     "cytoscape-avsdf": "^1.0.0",
     "cytoscape-cola": "^2.5.0",
     "deep-object-diff": "^1.1.0",
-    "express": "^4.16.4",
+    "express": "^4.18.2",
```
danburonline, line 54: General question regarding version pinning in our package.json: Are we intentionally avoiding the practice of restricting updates to patch versions only? I've observed that some packages, despite adhering to semantic versioning, can introduce breaking changes in minor updates, even in the absence of explicit API changes. Do we have a policy or consensus on managing such dependencies to mitigate potential issues? Barring this consideration, everything else looks good, and we can proceed with the merge once all checks are successful.

danburonline (approved):

package.json
bilalesi: interesting, can you give some examples of some packages that has breaking changes in the minor or patch versions ?

danburonline: Not specifically in this codebase, but based on experience. Here's some reading:
- https://maxleiter.com/blog/pin-dependencies
- https://the-guild.dev/blog/how-should-you-pin-dependencies-and-why

If we don't have any experimental components or complex proprietary libraries with specific peer dependencies, and if we haven't encountered any issues before, then it should be fine. I was just curious.

PR #1448, opened 2023-12-18
Optimize Elasticsearch view and indexing error retrieval

danburonline:
Fixes [#4552](https://github.com/BlueBrain/nexus/issues/4552)

## Description

pr

This pull request fixes over-etching for the indexing error UI. Previously, it would fetch all "failures" (indexing errors) when visiting, for example, this page: `http://localhost:8000/orgs/testing-org/daniel-example/settings`. This led to poor performance and potential application failure when a large number of indexing errors were returned simultaneously, even though the user did not see them in the UI.

The change that I've made is that it only fetches the indexing errors when expanding a view via the "+" icon. This improves the performance drastically (see screenshot above).

## How has this been tested?

- Dev and production build via my MacBook.

## Types of changes

## Checklist:


src/subapps/admin/components/Settings/ViewsSubView.tsx
```diff
@@ -227,7 +219,7 @@ const ViewsSubView = () => {
   const { mutateAsync: handleReindexingAllViews, isLoading } = useMutation(
     restartIndexingAllViews,
     {
-      onError: error => {
+      onError: () => {
```
bilalesi, line 222: it might be a good idea to capture the error in `Sentry`

src/subapps/admin/components/Settings/ViewsSubView.tsx
```diff
@@ -236,7 +228,56 @@ const ViewsSubView = () => {
...
+      Sentry.captureException(error, {
+        extra: {
+          orgLabel,
+          projectLabel,
+          viewId,
+          error,
+        },
+      });
+
+      return {
```
bilalesi, line 271: I don't think that returning the object when Error is a good idea, catching an error must not return a result object unless there is a good reason for that.
returning `null` is better

bilalesi:
Nice PR: one small UI wise improvement
when the expand the view and there is no errors, so either we show `0 Total Errors` or `Empy Icon 'No Data'`

Screenshot 2024-01-04 at 10 12 00

src/subapps/admin/components/Settings/ViewIndexingErrors.tsx
```diff
@@ -14,9 +14,9 @@ export const ViewIndexingErrors: React.FC<Props> = ({
   return (
     <div>
       {<h3>{indexingErrors._total} Total errors</h3>}
-      {indexingErrors._total && (
+      {indexingErrors._total > 0 && (
```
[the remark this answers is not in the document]
danburonline: I like it! Changed.

src/subapps/admin/components/Settings/ViewIndexingErrors.tsx
```diff
@@ -64,10 +64,10 @@ export const fetchIndexingErrors = async ({
 };
 
 export interface IndexingErrorResults {
-  '@context': string[] | string;
-  _next: string;
-  _results: IndexingError[];
+  '@context': any;
```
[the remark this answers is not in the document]
danburonline: That was a leftover from debugging. I reverted the types – thanks!

src/subapps/admin/components/Settings/ViewsSubView.tsx
danburonline: 👍

danburonline:
> Nice PR: one small UI wise improvement when the expand the view and there is no errors, so either we show `0 Total Errors` or `Empy Icon 'No Data'`
> 
> Screenshot 2024-01-04 at 10 12 00

Good idea, I will work on something 👍

src/subapps/admin/components/Settings/ViewsSubView.spec.tsx
```diff
@@ -75,26 +76,19 @@ describe('ViewsSubView', () => {
...
 
-    const indexingErrorRows = await getErrorRows();
-    expect(indexingErrorRows.length).toEqual(2);
-
-    const errorRow1 = await getErrorRow('Mock Error 1');
-    expect(errorRow1).toBeTruthy();
-    const errorRow2 = await getErrorRow('Mock Error 2');
-    expect(errorRow2).toBeTruthy();
+      const errorRow1 = getErrorRow('Mock Error 1');
+      expect(errorRow1).toBeTruthy();
```
Dinika, line 88: Nice to have: A very trivial change that is technically not a part of this diff (so you have every right to ignore). Looking at this test, I think it can be improved. 

The `getErrorRow` function finds elements by text. Since it uses the `findBy*` query the existence of the element is [implicitly already asserted](https://testing-library.com/docs/queries/about/#types-of-queries) so, in theory, the `expect(row).toBeTruthy` assertion is not really needed.

If you prefer for the tests to be explicit and still check for the existence of the row, I think. `expect(row).toBeVisible()` is a better assertion than `toBeTruthy` because it more closely matches the real scenario (i.e. we don't really want this row to be just any JS truthy value, we want it to be a visible DOM element).

src/subapps/admin/components/Settings/ViewsSubView.tsx
```diff
@@ -236,7 +235,51 @@ const ViewsSubView = () => {
...
+        nexus,
+        apiEndpoint,
+        orgLabel,
+        projectLabel,
+        viewId,
+      });
+
+      return indexingErrors;
+    } catch (error) {
+      console.error('Error fetching indexing errors on demand', error);
```
Dinika, line 263: Just curious - Why is this console needed here? If it's for error tracking purposes, I think the Sentry exception below would suffice so we can remove this.

src/subapps/admin/components/Settings/ViewsSubView.tsx
```diff
@@ -392,42 +435,72 @@ const ViewsSubView = () => {
...
         </Row>
 
-        <Table<TViewType>
+        <Table<SubView>
           loading={status === 'loading'}
           className="views-table"
           rowClassName="view-item-row"
           columns={columns}
-          dataSource={views?.results}
+          dataSource={views?.results as SubView[]}
```
Dinika, line 443: I might have overlooked something, but I think this type cast is technically incorrect. The `fetchViewsList` function returns a promise of an object where the `results` is *not* `SubView[]`, instead it is `Omit<Subview, 'indexingErrors'[]` (correct me if I'm wrong). Also, the `views?.results` can be `undefined` too. Therefore, the type casting might cause some bugs. 

I  would actually recommend to improve the type definition of this function instead of the explicit type casting here and on line [428](https://github.com/BlueBrain/nexus-web/pull/1448/files#diff-ce1ad27afb9d059ed2ab2fb4c9cf0a54ce5db0da412a4170d2be28afc06b11fdR428). So something like this can be added to the `fetchViewsList` function:

```typescript
const fetchViewsList = async ({
  nexus,
  orgLabel,
  projectLabel,
}: {
  nexus: NexusClient;
  orgLabel: string;
  projectLabel: string;
}): Promise<{                     // <----------------- Change
  errors: PromisePoolError<Omit<SubView, 'indexingErrors'>, any>[]; // <----------------- Change
  results: Omit<SubView, 'indexingErrors'>[]; // <----------------- Change
}> => {
```

After this, the type casting on both the lines should be removed (which might cause some issues with `undefined` which should be dealt with according to the situation).

src/subapps/admin/components/Settings/ViewsSubView.tsx
```diff
@@ -392,42 +435,72 @@ const ViewsSubView = () => {
...
+              })
+                .then(indexingErrors => {
+                  // Update the state with the new errors
+                  setExpandedRows(prevExpandedRows => ({
+                    ...prevExpandedRows,
+                    [r.key]: indexingErrors as IndexingErrorResults,
+                  }));
+                })
+                .catch(error => {
+                  console.error(
```
Dinika, line 487: Same question here about why we need this console (I'm a bit picky with these console errors because they don't really help us debug issues in production since the error is only logged on user's browser's console that the developer doesn't have access to, and they add too much noise in the console when we are developing. If for some reason, the error tracing is required, we should instead use Sentry maybe?)

Dinika (changes requested):
When I tested this everything looked and worked fine. I do have a question though: why are 3 requests triggered to `/failures` endpoint when I expand a row?

About the rest of the comments - the only important one is the one with the type casting. The others are only suggestions.

src/subapps/admin/components/Settings/ViewsSubView.tsx
danburonline: I noticed console errors/logs scattered throughout. I believe this is legacy code or noise. In my experience, we always added a linter rule to display a warning when there is a console.log statement, as it provides no benefit. Therefore, I suggest removing it.

danburonline:
> When I tested this everything looked and worked fine. I do have a question though: why are 3 requests triggered to `/failures` endpoint when I expand a row?
> 
> About the rest of the comments - the only important one is [the one with the type casting](https://github.com/BlueBrain/nexus-web/pull/1448#discussion_r1441920772). The others are only suggestions.

I had a look at it, and it should be fixed now. Please re-test and let me know! 😁

Dinika (approved):

Dinika:
Tested the repeated network calls issues - it seems to be fixed :sparkles:

PR #1449, opened 2024-01-05
Upgrade packages and move all to ESM 

bilalesi:
- Fix:  use `tsPathAliases` plugin in `vite` instead of using config.
- Fix: clean some parts of the code.
- Fix: use `tsx` for server in `dev` mode instead of `ts-node`.
- Update: `esbuild` config .
- Update: add package `engines` and update `.nvmrc`.

Note: In the Build process you will see something like: 
```shell
"/*#__PURE__*/"

in "node_modules/antd/es/card/Card.js" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.
node_modules/antd/es/list/Item.js (81:2) A comment
```

This will not impact the build since `Rollup` is not damaging the code instead he is not recognizing the position of the `__PURE__` comment which help when the builder do tree-shaking.

Fixes #

## Types of changes

- [] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My change requires a change to the documentation.


src/authManager.tsx
```diff
@@ -120,20 +120,33 @@ export const setupUserSession = async (userManager: UserManager) => {
...
+  if (user) {
+    // we've loaded a user, refresh it
+    try {
       user = await userManager.signinSilent();
+    } catch (error) {
+      console.error(
+        '@@error',
+        "we've loaded a user, refresh it raised an error",
+        error
+      );
```
danburonline, line 138: Should we print anything to the console? Specifically, I'm wondering if it's necessary or beneficial to print things to the console in general. Does it provide any assistance or value to us or the user? Alternatively, why not use Sentry, e.g.?

package.json
```diff
@@ -1,18 +1,24 @@
 {
   "name": "nexus-fusion",
   "description": "The interface of Blue Brain Nexus, the open-source knowledge graph for data-driven science.",
-  "version": "1.9.1",
+  "version": "1.10.x",
```
danburonline, line 4: Shouldn't we avoid defining any version number at all since we have to do it manually and might forget? After all, it's not a package deployed on NPM, so the version number is not crucial.

package.json
```diff
@@ -1,18 +1,24 @@
...
+    "yarn": "1.22.x"
+  },
   "scripts": {
     "clean": "rm -rf ./dist",
     "dev:client": "yarn clean && concurrently 'vite build -c ./vite.config.refresh.js' 'vite'",
-    "dev:server": "NODE_ENV=development ts-node-dev --project tsconfig.node.json -- ./server/index.ts",
+    "dev:server": "NODE_ENV=development tsx --tsconfig ./tsconfig.node.json ./server/index.ts",
     "dev": "concurrently 'yarn dev:client' 'yarn dev:server'",
-    "build:server": "etsc --config=etsc.config.js",
+    "build:server": "etsc --config=etsc.config.cjs",
```
danburonline, line 21: Question: Why did you change the config to a CommonJS file? Was there a particular reason for it?

package.json
bilalesi: yeah you are right, I ll remove it

src/authManager.tsx
bilalesi: absolutely right, change it 👍

danburonline (approved):

PR #1466, opened 2024-01-10
Generate manifest dynamically

bilalesi:
1. use `new URL('.', import.meta.url).pathname;` to get `__dirname` instead of `path/url` modules.
2. generate `favicon.ico` for diffent base paths ex: `/nexus`.
3. generate `manifest.json` dynamically for different `base_url`.
4. fix fn declaration name `getPreloadedState`.

## Types of changes

## Checklist:


index.html
```diff
@@ -4,10 +4,7 @@
 <head>
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-  <link rel="icon" href="/public/favicon.ico" sizes="48x48" />
-  <link rel="icon" href="/public/favicon.svg" sizes="any" type="image/svg+xml" />
-  <link rel="apple-touch-icon" href="/public/apple-touch-icon-180x180.png" />
-  <link rel="manifest" href="/public/manifest.json" />
+  <!--app-manifest-->
```
danburonline, line 7: Maybe you could add a comment here to say that the manifest will be dynamically generated and inserted here? Otherwise, this comment may be confusing.

danburonline (approved):

index.html
bilalesi: Actually this is not a comment to describe the use, this is placeholder to replace it when the server generate the right `link` tags

PR #1468, opened 2024-01-10
Undeprecate other resource types (`schemas`, `views` and `storages`)

danburonline:
Fixes [#4586](https://github.com/BlueBrain/nexus/issues/4586)

This PR enables undoing the deprecation of other resource types: schemas, views and storages. `Views` also means types such as `SparqlView`, `CompositeView,` and `ElasticSearchView`.

## Description

- Removed the part that hid the "undo deprecation" button from unsupported but now supported types.
- Extended the function that constructs the undeprecation endpoint.

## How has this been tested?

- Tested the undeprecation of various types of resources via the local dev server and prod build. 

## Types of changes

### DIFF
--- a/src/shared/containers/ResourceViewContainer.tsx
+++ b/src/shared/containers/ResourceViewContainer.tsx
@@ -74,7 +74,10 @@ function constructUnDeprecateUrl(
         return 'files';
       case 'Storage':
         return 'storages';
-      case 'View' || 'ElasticSearchView' || 'SparqlView':
+      case 'ElasticSearchView':
+      case 'SparqlView':
+      case 'CompositeView':
+      case 'View':
         return 'views';
       case 'Schema':
         return 'schemas';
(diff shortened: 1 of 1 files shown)


src/shared/containers/ResourceViewContainer.tsx
```diff
@@ -67,14 +67,46 @@ function constructUnDeprecateUrl(
...
+
+  const primaryResourceType = resource['@type'] || '';
+  const resourcePathSegment = determineResourcePathSegment(primaryResourceType);
+  const slashPrefix = Array.isArray(primaryResourceType)
+    ? primaryResourceType.some(type =>
+        Object.keys(typePathMapping).includes(type)
+      )
+      ? ''
+      : '_/'
+    : Object.keys(typePathMapping).includes(primaryResourceType)
```
Dinika, line 103: Nice to have: To keep it consistent with line 87 above should we use `primaryResourceType in typePathMapping` here? It is also slightly better in terms of memory consumption than creating a new array with `Object.keys`

Dinika (approved):

PR #1469, opened 2024-01-10
update: use https for all env except local machine

bilalesi:
## Types of changes

## Checklist:

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -185,7 +185,9 @@ app.get(`${base}/status`, (_, res) => {
 });
 
 app.get(`${base}/web-manifest`, (req, res) => {
-  const startUrl = `${req.protocol}://${req.get('host')}${base}`;
+  const host = req.get('host');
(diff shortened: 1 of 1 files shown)


server/index.ts
```diff
@@ -185,7 +185,9 @@ app.get(`${base}/status`, (_, res) => {
 });
 
 app.get(`${base}/web-manifest`, (req, res) => {
-  const startUrl = `${req.protocol}://${req.get('host')}${base}`;
+  const host = req.get('host');
+  const isLocalhost = host?.startsWith('localhost');
+  const startUrl = `${isLocalhost ? req.protocol : 'https'}://${host}${base}`;
```
Dinika, line 190: Only a suggestion - should we use here `process.env.NODE_ENV` to check it's a development environment then use `http`?

Dinika (approved):

server/index.ts
bilalesi: good point 👍

PR #1471, opened 2024-01-12
F4650 fix duplicate editor project query tab

bilalesi:
1. Fix manifest loading in `dev` mode (localhost).
2. Fix duplicate Codemirror editor in SparQL Query tab (project page).
3. Fix duplicate Codemirror editor in ES Query tab (project page)

## Types of changes

## Checklist:


src/subapps/admin/components/ViewForm/SparqlQueryInput.tsx
```diff
@@ -1,25 +1,30 @@
-import React from 'react';
+import React, { useRef } from 'react';
+import * as codemirror from 'codemirror';
+import { UnControlled as UnControlledCodeMirror } from 'react-codemirror2';
```
danburonline, line 3: General understanding question: Why do we need codemirrror and react-codemirror2?

danburonline (approved):

src/subapps/admin/components/ViewForm/SparqlQueryInput.tsx
bilalesi: `react-codemirror2` require `codemirror`, we can use only codemirror, just was used long time before and also in the same time it's very helpful to do all the config abstraction

PR #1472, opened 2024-01-12
F4643 project query tab style

bilalesi:
1. Fix some bad use of `useEffect` both in ES/SparQL query tabs. (use events instead)
2. use `useQuery/useMutation` instead of `useEffect` (bad use of useEffect on mutation the query editor).
3. Fix style of query editor and query results (remove duplicate pagination components).

**NOTE** both components need refactoring in next part of migration.

## Types of changes

## Checklist:

### DIFF
--- a/src/subapps/admin/containers/ElasticSearchQuery.tsx
+++ b/src/subapps/admin/containers/ElasticSearchQuery.tsx
@@ -1,8 +1,14 @@
-import * as React from 'react';
+import { useState, FC } from 'react';
 import { useNexusContext } from '@bbp/react-nexus';
-import { DEFAULT_ELASTIC_SEARCH_VIEW_ID } from '@bbp/nexus-sdk/es';
+import {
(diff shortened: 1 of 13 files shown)


src/subapps/admin/containers/ElasticSearchQuery.tsx
```diff
@@ -73,29 +61,28 @@ const ElasticSearchQueryContainer: React.FunctionComponent<{
...
-      response={response}
-      busy={busy}
-      error={error}
-      from={from}
-      size={size}
-      onPaginationChange={onPaginationChange}
-      onQueryChange={onQueryChange}
-      onChangePageSize={onChangePageSize}
-    />
+    <>
```
danburonline, line 73: Why is a Fragment needed here? It's just one child element?

bilalesi: Missed it, thanks

danburonline (approved):

PR #1474, opened 2024-01-17
Adjust custom linter logic for code editor

danburonline:
## Description

It was mentioned on Slack that only top-level fields of JSON-LD should have a constraint in the code editor component. This constraint prevents users from starting field names with an underscore. The previous implementation disallowed adding an underscore at the beginning of every field, rather than just the top-level ones. The change in this PR removes this feature and disallows it only for the most top-level ones.

## How has this been tested?

- Prod and dev builds locally.

## Types of changes


bilalesi:
nice work, one new to have in tests is to test for non-Top-level properties 👍

danburonline:
> nice work, one nice to have in tests is to test for non-Top-level properties 👍

I added an additional test for that 👍

bilalesi (approved):

bilalesi:
LGTM 👍

PR #1495, opened 2024-02-19
Don't show linter issues when `showMetadata` toggling

danburonline:
Issue found by Bilal. Fixing the problem where the linter issues are not shown when the `showMetaData` toggle is set to true, as it then displays linter errors caused by the underscores in the metadata.

Also:
- I removed the "You can edit this resource" information as I actually removed this in a PR before the migration
- I changed and cleaned some styles that I used to change and adjust in the original PR #1443

## How has this been tested?

- Local dev and prod builds pointing to `staging`.

## Types of changes

### DIFF
--- a/src/shared/containers/AdminPluginContainer.tsx
+++ b/src/shared/containers/AdminPluginContainer.tsx
@@ -71,23 +71,6 @@ const AdminPlugin: React.FunctionComponent<AdminProps> = ({
 
   const content = (
     <>
-      <AccessControl
-        path={`/${orgLabel}/${projectLabel}`}
-        permissions={['resources/write']}
-        noAccessComponent={() => <></>}
-      >
-        {editable && (
-          <Alert
-            message={
-              <>
-                <EditOutlined /> You can edit this resource.
-              </>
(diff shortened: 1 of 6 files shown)


src/shared/containers/AdminPluginContainer.tsx
```diff
@@ -71,23 +71,6 @@ const AdminPlugin: React.FunctionComponent<AdminProps> = ({
...
-            message={
-              <>
-                <EditOutlined /> You can edit this resource.
-              </>
-            }
-            type="success"
-          />
-        )}
-      </AccessControl>
-
```
bilalesi, line 90: is there is a reason why you remove the check for permission for this one

danburonline: `I removed the "You can edit this resource" information as I actually removed this in a PR before the migration`. We once talked about this if I could remove this some months ago, and it came back after the migration rebase => so I removed it again. I can bring it back if you think it's still necessary.

bilalesi (approved):

PR #1498, opened 2024-02-20
fix: omit the resource metadata from the update endpoint

bilalesi:
fix update studio details

reproduction: 
1. select studio
2. click on edit studio
3. try to update at least one field as title, or description

fix: the payload of the update resource endpoint must only contains the relative data without the metadata 
either way delta will fail

extra: I update the description of the studio in the studio listing view to show only 2 lines of the description, I noticed that many studio has a big description 

## Types of changes

## Checklist:


Dinika (approved):
Tested both changes. Works well.

PR #1499, opened 2024-02-20
Fix/migration/studio update

bilalesi:
Fix studio update
Fix resource actions container

## Types of changes

## Checklist:


src/subapps/studioLegacy/containers/StudioContainer.tsx
```diff
@@ -114,7 +116,7 @@ const StudioContainer: React.FunctionComponent = () => {
         studioId,
         studioResource._rev,
         {
-          ...studioResource,
+          ...omitBy(studioResource, (_, key) => key.trim().startsWith('_')),
```
danburonline, line 119: General question: Given the advancements in modern JavaScript, I'm wondering if we could lean more on native methods like filter and reduce for this case, rather than relying on lodash?

danburonline (approved):
LGTM. Just one question left.

PR #1505, opened 2024-02-21
Download of resources with tags in their ids should work

Dinika:
When storing these resources and their distributions to the localstorage the tag should be stored as well. Otherwise the `key` (required to add/remove items from data panel) is inconsistent, leading to bugs (example, adding the resource works, but removing them from the data panel does not, because the id is not the same).

Fixes #

## Types of changes

## Checklist:


bilalesi:
LGTM, just need cleaning the logs.

bilalesi (approved):

PR #1540, opened 2024-07-02
DO_NOT_MERGE: use GET query instead of post with only type as path param

bilalesi:
**IMPORTANT** do not merge


src/shared/utils/querySparqlView.ts
```diff
@@ -26,25 +26,53 @@ export const sparqlQueryExecutor = async (
...
+    `${apiEndpoint}/views/${orgLabel}/${projectLabel}/${type}`,
+    {
+      headers: {
+        Accept: 'application/json',
+        Authorization: `Bearer ${localStorage.getItem('nexus__token')}`,
+      },
+    }
+  )
+    .then(res => res.json())
+    .catch(er => console.error);
```
bilalesi, line 60: this is the customization of the `sparql query`

